PR #1, opened 2023-09-27
Develop

quadrismegistus:


quadrismegistus:
Wow, thanks for unceremoniously deleting all of my work here, much of it useful, which I explicitly asked to be kept around, like Vineet's script was. But when have my requests ever been listened to? Removing myself from this repository and conversation since I am still getting notifications

rlskoeser:
@quadrismegistus sorry it looks like it was deleted; I see that it certainly looks that way from the GitHub notification. I didn't delete it, I just renamed the branch while we figure out how to restructure the code so we can pull in the relevant parts when we're ready.

quadrismegistus:
Thanks for explaining. Surprised I still have access to this repository since I had to beg for access to geotaste. Removing myself from notifications regardless

PR #7, opened 2024-03-25
initial setup for corppa python package

rlskoeser:
- update readme with basic preliminary content
- add pyproject.toml to configure as an installable python package; pull version from package
- setup and document pre-commit with ruff for python linting and formatting


README.md
```diff
@@ -1,2 +1,21 @@
...
+
+This repo uses [git-flow](https://github.com/nvie/gitflow) branching conventions; **main** contains the most recent release, and work in progress will be on the **develop** branch. Pull requests for new features should be made against develop.
+
+### Developer setup and installation
+
+- **Recommended:** create a python virtual environment with your tool of choice (virtualenv, conda, etc); use python 3.10 or higher
+
+- Install the local checked out version of this package in editable mode (`-e`), including all python dependencies  and optional dependencies for development and testing:
+```sh
+pip install -e .[dev]
```
laurejt, line 15: Add the quotes so that this will play nicely with both macOS and Windows

```suggestion
pip install -e ".[dev]"
```

laurejt (changes requested):
Seems good to me overall, just need to make the one instruction change. Note that since this is an initial setup, I didn't look too closely at the contents of the `.toml` or `.yaml` files.

README.md
rlskoeser: oh, good catch! yeah, that's actually necessary in my shell too

laurejt (approved):

Issue #5, opened 2024-03-19
As a researcher, I want to be able to get the text corpus for a subset of record IDs so that I can conduct textual analysis within particular groups of texts.


mnaydan:
## acceptance criteria
- [x] filter script successfully filters the corpus and generates a subset corpus that includes all pages for the specified source ids and only pages for those source ids
    - [x]  for simple `idfile` (one source_id per line; non-excerpt sources)
    - [x] for simple `idfile` with leading whitespace
    - [x] for simple `idfile` with trailing whitespace
    - [x] for `idfiles` with sources with 1 or more excerpts
    - [x] for `idfiles` with a mix of existing and non-existing sources
- [x] filter script has reasonable error handling (missing/empty id file, output file already exists)

rlskoeser:
Here's a screen recording of my terminal showing the corpus subset functionality:


https://github.com/Princeton-CDH/ppa-nlp/assets/691231/68057e2c-e324-468f-96e5-7fde847a6549

laurejt:
Should an output file be created if there are no matches?

rlskoeser:
> Should an output file be created if there are no matches?

I guess ideally not, but I don't think there's an easy way to detect this until after consuming the generator. I guess we could check and remove it after if it's zero size.

laurejt:
Testing acceptance complete. The only outstanding issue, is whether the script should produce empty output files when no matches are found (either way seems reasonable).

rlskoeser:
Thanks for testing. Let's leave the empty file behavior as it is for now, we've both put enough time into this feature already. We can always revisit and tweak it later if we find it's causing problems.

PR #8, opened 2024-03-28
Functionality to filter/subset PPA full-text corpus by source id

rlskoeser:
ref #5

### DIFF
--- a/src/corppa/utils/filter.py
+++ b/src/corppa/utils/filter.py
@@ -1,10 +1,27 @@
+"""
+Utility for filtering PPA full-text corpus to work with a subset of
+pages. Currently supports filtering by a list of PPA source ids.
+
+Can be run via command-line or python code. Takes jsonl file (compressed or
+not) as input, a filename for output, and a file with a list of
+selected source ids.
+
+To use as a command-line script:
+```
+corppa-filter-corpus path/to/ppa_pages.jsonl output/ppa_subset_pages.jsonl my_ids.txt
+```
+Input format and output filename can use any extension supported by mod:`orjsonl`,
+with or without compression; e.g. `.jsonl`, `.jsonl.gz`, `.jsonl.bz2`, etc.
+
+"""
+
 import argparse
 
 import orjsonl
 from tqdm import tqdm
 
 
-def filter_pages(input_filename, source_ids):
+def filter_pages(input_filename, source_ids, disable_progress=False):
     """Takes a filename for a PPA full-text corpus in a format orjsonl supports
     and a list of source ids. Returns a generator of filtered pages from the
     full corpus corresponding to the list of ids.
@@ -17,6 +34,7 @@ def filter_pages(input_filename, source_ids):
         orjsonl.stream(input_filename),
         desc="Filtering",
         bar_format="{desc}: {n:,} pages{postfix} | elapsed: {elapsed}",
+        disable=disable_progress,
     )
     for page in progress_pages:
         # page data does not include source id, but does include work id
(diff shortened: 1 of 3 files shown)


src/corppa/utils/filter.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def save_filtered_corpus(input_filename, output_filename, idfile):
+    """Takes a filename for input PPA full-text corpus in a format
+    orjsonl supports, filename where filtered corpus should be saved,
+    and a filename with a list of source ids, one id per line.
+    """
+    # read the id file and generate a list of ids
+    with open(idfile) as idfile_content:
+        source_ids = [line.strip() for line in idfile_content.readlines()]
```
laurejt, line 61:
```suggestion
        source_ids = [line.strip() for line in idfile_content]
```
```

File: src/corppa/utils/filter.py [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+        description="Filters PPA full-text corpus by list of source ids",
+    )
+    parser.add_argument(
+        "input",
+        help="path to PPA full-text corpus to be "
+        + "filtered; must be a JSONL file (compressed or not)",
+    )
+    parser.add_argument(
+        "output", help="filename where the filtered corpus should be saved"
+    )
```
laurejt, line 79: What kind of behavior is desired for missing/inappropriate file extensions? The script will work, producing a json file. Perhaps ".json" should be appended to the filename in these cases?

laurejt (changes requested):

src/corppa/utils/filter.py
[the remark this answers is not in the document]
rlskoeser: I think I had that at one point and then I took it out! I wasn't sure if we want both ways to run it - if you install with pip / pyproject.toml it can be run with the name defined there.   But it doesn't hurt to have options; I'll add this back in.

laurejt (approved):

rlskoeser:
@laurejt I pushed another set of changes with a bit more documentation and error handling for the script/main function.  I added type hints in a few places because I'd like to use a python docs solution that uses them.  I switched the parameter order - when I was demoing the script, it seemed more logical to me that the inputs should come before the desired output filename.

This is ready for you to test now. Would it be useful for me to draft a testing checklist on #5 ?

rlskoeser:
ha, belay that 😅 - messed myself up with the argument order change, adding a bit more error checking

rlskoeser:
OK! Now updated with better error handling on the files and corresponding updated unit tests.  

I started a testing checklist on https://github.com/Princeton-CDH/ppa-nlp/issues/5

src/corppa/utils/filter.py
```diff
@@ -97,13 +123,33 @@ def main():
...
+    if not os.path.exists(args.idfile):
+        print(f"Error: idfile {args.idfile} does not exist")
+        exit(-1)
+    elif os.path.getsize(args.idfile) == 0:
+        print(f"Error: idfile {args.idfile} is empty")
+        exit(-1)
+
     # if requested output filename has no extension, add jsonl
+    output_filename = args.output
     if os.path.splitext(output_filename)[1] == "":
```
laurejt, line 135: This is assuming an "extension" simply corresponds to the existence of a `.` in the filename. This is probably fine, but will produce `.jsonl` files without the appropriate extension in some cases.

src/corppa/utils/filter.py
```diff
@@ -97,13 +123,33 @@ def main():
...
+
+    try:
+        save_filtered_corpus(
+            args.input, args.idfile, output_filename, disable_progress=disable_progress
+        )
+    except (FileNotFoundError, JSONDecodeError) as err:
+        # catch known possible errors and display briefly
+        # with the type of error and the brief message
+        print(f"{err.__class__.__name__}: {err}")
+        exit(-1)
```
laurejt, line 152: See comment for Line 128

```suggestion
        sys.exit(-1)
```

src/corppa/utils/filter.py
```diff
@@ -97,13 +123,33 @@ def main():
     # progress bar is enabled by default; disable if requested
     disable_progress = not args.progress
 
-    output_filename = args.output
+    if not os.path.exists(args.idfile):
+        print(f"Error: idfile {args.idfile} does not exist")
+        exit(-1)
```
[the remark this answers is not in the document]
rlskoeser: I did not know that, thank you for flagging and sharing the link!

src/corppa/utils/filter.py
```diff
@@ -97,13 +123,33 @@ def main():
...
+    elif os.path.getsize(args.idfile) == 0:
```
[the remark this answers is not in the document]
rlskoeser: yes, true - I thought this was reasonable for a simple check before processing (this was after I accidentally obliterated my id text file and replaced it with an empty file 😂 )

src/corppa/utils/filter.py
rlskoeser: I don't think we should worry about that now as it seems unlikely, but if you have specific cases in mind we could document in case we want to revisit.

rlskoeser:
@laurejt thanks for the corrections and improvements. sys.exit is also much nicer to test, fwiw (I'm now using `pytest.raises(SystemExit)`). I've incorporated your changes and updated the unit tests.

PR #30, opened 2024-05-21
Scripts update & cleanup

laurejt:
This feature includes a new helper method for obtaining the "clean" version of a HathiTrust volume identifier, as well as a general reorganization of the experimental scripts directory.

### DIFF
--- a/scripts/helper.py
+++ b/scripts/helper.py
@@ -8,18 +8,33 @@ import gzip
 import bz2
 
 
-_char_conversion_map = {"ſ": "s"}
-_char_translation_table = str.maketrans(_char_conversion_map)
+_htid_encode_map = {":": "+", "/": "=", ".": ","}
+_htid_encode_table = str.maketrans(_htid_encode_map)
+_htid_decode_map = {v: k for k, v in _htid_encode_map.items()}
+_htid_decode_table = str.maketrans(_htid_decode_map)
 
 
-def clean_htid(htid):
+def encode_htid(htid):
     """
-    Returns the "clean" version of a HathiTrust volume id
+    Returns the "clean" version of a HathiTrust volume identifier with the form:
+        [library id].[volume id]
+    Specifically, the volume-portion of the id undergoes the following
+    character replacement: ":" --> "+", "/" --> "=", "." --> ","
     """
     lib_id, vol_id = htid.split(".", 1)
-    vol_id = vol_id.replace(":", "+")
-    vol_id = vol_id.replace("/", "=")
-    vol_id = vol_id.replace(".", ",")
+    vol_id = vol_id.translate(_htid_encode_table)
+    return f"{lib_id}.{vol_id}"
+
+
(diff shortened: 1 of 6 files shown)


scripts/helper.py
```diff
@@ -1,20 +1,28 @@
...
+def clean_htid(htid):
+    """
+    Returns the "clean" version of a HathiTrust volume id
+    """
+    lib_id, vol_id = htid.split(".", 1)
+    vol_id = vol_id.replace(":", "+")
+    vol_id = vol_id.replace("/", "=")
+    vol_id = vol_id.replace(".", ",")
+    return f"{lib_id}.{vol_id}"
+
```
rlskoeser, line 24: Is this a standard conversion? I'm not as familiar with the HTRC featureset code & approaches; was glancing at that and see how this is similar (or same?). Would benefit from slightly more context in the method explaining what the clean version is used for.

Is it worth using `translate` here instead of `replace` like you did for the other character replacement?

rlskoeser (approved):
Thanks for reorganizing the script documentation. this is an improvement.

I think the use case for the clean htid could be clarified a bit, but otherwise all of this looks good.

I think some of these utilities / helpers should be part of the `corppa` package rather than one-off utility code, but I guess we don't have enough information yet to know which pieces will be reusable.

scripts/helper.py
laurejt: This is the conversion that HTRC has used for extracted features (see this [utility function](https://github.com/htrc/htrc-feature-reader/blob/master/htrc_features/utils.py#L9)); it's also documented [here](https://htrc.atlassian.net/wiki/spaces/COM/pages/43296152/Stubbytree+directory+structure).

I can add the text that explains the conversion. It really is just remapping a few characters.

I can make the change to `translate`, since it's easy if not strictly necessary. It's probably slightly faster, although this isn't a place where we need to be concerned about performance.

laurejt:
> Thanks for reorganizing the script documentation. this is an improvement.
> 
> I think the use case for the clean htid could be clarified a bit, but otherwise all of this looks good.
> 
> I think some of these utilities / helpers should be part of the `corppa` package rather than one-off utility code, but I guess we don't have enough information yet to know which pieces will be reusable.

I completely agree that several of these scripts should be incorporated to the `corppa` package at some point, mostly just a matter of when it's worth doing since then we may need to rethink dependencies / environments.

scripts/helper.py
rlskoeser: I had the same thought about `translate` - not strictly necessary.

So the clean id is for use with stubby tree / filenames? Are there other use cases?

Linking to the documentation is probably enough.

rlskoeser (approved):
great!

scripts/helper.py
laurejt: Just to leave a record, this is really just following the cleaning procedure of pairtree / more generally making sanitizing htids so that filenames play nice with the filesystem.

PR #40, opened 2024-07-02
Added helper function for stub directory logic

laurejt:


scripts/helper.py
```diff
@@ -8,6 +8,19 @@
...
 
+def get_stub_dir(source, vol_id):
+    """
+    Returns the stub directory for the specified volume (vol_id) and
+    source type (source)
+    """
+    if source == "Gale":
+        return vol_id[::3][1:]
+    elif source == "HathiTrust":
+        return vol_id.split(".", maxsplit=1)[0]
```
rlskoeser, line 19: would be helpful to have a one line comment or example for each of these

logic looks good

laurejt: How's it look now?

rlskoeser: looks great!

rlskoeser (approved):

PR #42, opened 2024-07-02
Revise filter script with options to include/exclude based on page attributes

rlskoeser:
Update filter script to make filter by id optional and add flags to include / exclude pages based on arbitrary attributes. 

I successfully used this to filter the poetry test set into poetry and non-poetry chunks based on the `contains_poetry` flag that was set by Laure's `create_pageset` script.

## changes:
- [x] update docstring for `filter_pages` to document new params
- [x] throw error if no filter args are specified
- [x] check filter logic: can we continue on failure

### DIFF
--- a/src/corppa/utils/filter.py
+++ b/src/corppa/utils/filter.py
@@ -34,7 +34,7 @@ from tqdm import tqdm
 
 def filter_pages(
     input_filename: str,
-    source_ids: list[str] | None,
+    source_ids: list[str] | None = None,
     disable_progress: bool = False,
     include_filter: dict | None = None,
     exclude_filter: dict | None = None,
@@ -144,6 +144,8 @@ class MergeKeyValuePairs(argparse.Action):
     """
 
     # adapted from https://stackoverflow.com/a/77148515/9706217
+
+    # NOTE: we may want a multidict here to support multiple values for the same key
 
     def __call__(self, parser, args, values, option_string=None):
         previous = getattr(args, self.dest, None) or dict()
--- a/src/corppa/utils/filter.py
+++ b/src/corppa/utils/filter.py
@@ -179,12 +179,22 @@ def main():
         "output", help="filename where the filtered corpus should be saved"
     )
     parser.add_argument(
+        "--progress",
+        help="Show progress",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+    )
+    filter_args = parser.add_argument_group(
+        "filters",
+        "Options for filtering pages. At least one filter is required",
+    )
+    filter_args.add_argument(
         "-i",
         "--idfile",
         help="filename with list of source ids, one per line",
         required=False,
     )
-    parser.add_argument(
+    filter_args.add_argument(
         "--include",
         nargs="*",
         action=MergeKeyValuePairs,
@@ -192,7 +202,7 @@ def main():
         help='Include pages by attribute: add key-value pairs as key=value or key="another value". '
         + "(no spaces around =, use quotes for values with spaces)",
     )
-    parser.add_argument(
+    filter_args.add_argument(
         "--exclude",
         nargs="*",
         action=MergeKeyValuePairs,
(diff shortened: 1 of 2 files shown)


test/test_utils/test_filter.py
```diff
@@ -33,6 +33,47 @@ def test_filter_pages(corpus_file):
...
+    results = list(
+        filter_pages(
+            str(corpus_file),
+            disable_progress=True,
+            include_filter={"work_id": "bar-p1", "label": "23"},
+        )
+    )
+    assert len(results) == 4
+    assert set([r["work_id"].split("-")[0] for r in results]) == {"bar", "baz"}
+    assert set([r["label"] for r in results]) == {"1", "2", "3", "23"}
```
laurejt, line 59: This test does not test the `filter_exclude` parameter.

test/test_utils/test_filter.py
```diff
@@ -33,6 +33,47 @@ def test_filter_pages(corpus_file):
...
+
+
+def test_filter_exclude(corpus_file):
+    results = list(
+        filter_pages(
+            str(corpus_file),
+            disable_progress=True,
+            include_filter={"work_id": "bar-p1", "label": "23"},
```
laurejt, line 54:
```suggestion
            exclude_filter={"work_id": "bar-p1", "label": "23"},
```

test/test_utils/test_filter.py
laurejt, line 59:
```suggestion
    assert len(results) == 1
    assert set([r["work_id"].split("-")[0] for r in results]) == {"foo"}
    assert set([r["label"] for r in results]) == {"i"}
```

src/corppa/utils/filter.py
```diff
@@ -59,42 +64,99 @@ def filter_pages(
...
+
+
+class MergeKeyValuePairs(argparse.Action):
+    """
+    custom argparse action to split a KEY=VALUE argument and append the pairs to a dictionary.
+    """
+
+    # adapted from https://stackoverflow.com/a/77148515/9706217
+
+    # NOTE: we may want a multidict here to support multiple values for the same key
```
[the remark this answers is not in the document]
rlskoeser: I like that approach but think we should hold off on implementing until we have a use case.

Should I add / revise the comment in the code so we don't lose track?

src/corppa/utils/filter.py
```diff
@@ -59,42 +64,99 @@ def filter_pages(
...
+                not any(page[key] == val for key, val in exclude_filter.items())
+            )
+
+        # make sure we have at least one True flag and all flags are True
+        if include_page and all(include_page):
             # keep track of how many have been selected for reporting in
             # progress bar
             selected_pages += 1
             progress_pages.set_postfix_str(f"selected {selected_pages:,}")
             yield page
```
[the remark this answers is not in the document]
rlskoeser: Thanks for catching that. I've refactored and the tests pass, but please review to confirm the logic.

src/corppa/utils/filter.py
```diff
@@ -40,17 +40,27 @@ def filter_pages(
...
-    and a list of source ids. Returns a generator of filtered pages from the
-    full corpus corresponding to the list of ids.  Displays progress
-    with :mod:`tqdm` progress bar unless disabled.
+    and one or more options for filtering that corpus. Returns a generator of
+    filtered pages from the full corpus corresponding to the list of ids.
+    At least one filtering option must be specified.
+    Displays progress with :mod:`tqdm` progress bar unless disabled.
 
     :param input_filename: str, filename for corpus input
     :param source_ids: list of str, source ids to include in filtered pages
```
laurejt, line 49: Is the parameter`source_ids` required or optional? If it's optional, add this to the parameter description.

src/corppa/utils/filter.py
```diff
@@ -40,17 +40,27 @@ def filter_pages(
...
+    :param include_filter: dict of key-value pairs for pages to include in
+        the filtered page set; equality check against page data attributes (optional)
+    :param exclude_filter: dict of key-value pairs for pages to exclude from
+        the filtered page set; equality check against page data attributes (optional)
     :param disable_progress: boolean, disable progress bar (optional, default: False)
     :returns: generator of dict with page data
     :raises: FileNotFoundError, orjson.JSONDecodeError
     """
     # convert list of source ids to set for fast hashmap lookup
```
laurejt, line 58: Relocate this comment to associated code (above L64). Consider adding the following comment for the input parameter check, since it's not documented in the method-level comment (taken from `save_filtered_corpus`)

laurejt:
```suggestion
    # at least one filter is required
```

laurejt (commented):
Looks good. Update the documentation for `filter_pages` before merging.

Issue #39, opened 2024-07-02
Code and Prodigy configuration for poem annotation task


rlskoeser:
Adapt from Wouter's recipe and prep code shared in Google Drive; code will be added to ppa-nlp/corppa

- [x] update ansible playbook to serve out images for annotation
- [x] add utility script to add metadata to page data for prodigy
- [x] add custom css to display image+text side by side

# separate but related tasks
- #46 
- reconciling the different data prep scripts (create_pageset, filter, add_metadata) 
- scripted / replicable solution for selecting a specified number of random pages from a jsonl corpus

rlskoeser:
progress so far:
- added Wouter's code to a branch with a co-author commit as a starting point
- updated prodigy ansible playbook to serve out ppa images from tigerdata (ran into permissions problems) and proxy the prodigy app running on port 8080
- requested PUL nginxplus config change to proxy port 80 instead of 8080 so it will get both images and prodigy app
- using Laure's version of the poem test set pages that I filtered to just the pages that contain pages, for easier testing
- added custom css file and configured prodigy recipe to include it; using css grid to display image and text side by side
- running prodigy with the recipe from a branch of corppa code

still to do:
- clean up / reorganize the code
- adapt and use Wouter's logic to update the jsonl file to add work-level metadata where Prodigy expects it
  -  make sure css customization works with Prodigy metadata block

blockers:
- I'm getting an error trying to save annotations on the images and text - they work separately but not together; I opened a [support request on the Prodigy Support forum](https://support.prodi.gy/t/errors-trying-to-use-image-manual-and-spans-manual-at-the-same-time-typeerror-on-tokens-token-start-and-points/7290)

rlskoeser:
Here's the mogrify command I was using for bulk convert + resize of the Gale images:
`mogrify -format jpg -resize 500 */*.TIF`

We probably want to resize the HT images as well, I think the versions we have now are quite a bit larger than we need for this interface.

rlskoeser:
I've updated this so the tasks we're putting off and/or need to discuss and make decisions about are separate tasks, but the code for this is all in branches that I think should be reviewed, so I'm putting it under review.

rlskoeser:
Using command line tools to make a test set of pages for annotation (or at least training on) with a 176 pages from HT known to contain poetry and 200 other pages from the same volume.

[...]

PR #43, opened 2024-07-03
Prodigy poetry annotation recipes and data prep

rlskoeser:
ref #39 

included in this PR:
- custom recipes for Prodigy annotation
- CSS for two-column page layout in Prodigy recipes
- script to add work-level metadata to page jsonl for display in Prodigy

## revisions
- [x] add top-level docstring to add metadata script
- [x] clean up corppa-?? comment (add note about refactor)
- [x] add comments to clarify image path variable is a base url
- [x] clean up outdated comments  about usage


src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -0,0 +1,172 @@
...
+        "labels": {
+            # trying to use options from PPA webapp color scheme,
+            # but may not be so great in Prodigy UI.
+            # azure #0788fc seafoam blue #57c4c4 wisteria #9c93c0 pig pink #ed949c
+            "POETRY": "#57c4c4",  # label color for POETRY
+        },
+        "hide_true_newline_tokens": False,
+    },
+    "global_css_dir": CURRENT_DIR,
+}
```
laurejt, line 40: Some additional settings we may want:
- `"show_whitespace": true`
- `"instructions": [text file, perhaps excerpt from handbook?]`
- `"allow_work_stealing": false` to ensure that annotators can't "steal" each others loads (hopefully not an issue when they both need to annotate everything)

For annotation assignments we should set either (I believe these settings are meant to be used with sessions):
1. `"feed_overlap": true` so that every session should annotate *all* pages*, or
2. `"annotations_per_task": 2` so that every page gets at least 2 annotations (although I'm not sure allocation works in this case)
I think we want (1) at the moment, but (2) could become useful if we have more annotators (and don't want/need them to annotate the same things)

laurejt: On batch & history sizes, by default it's 10. I think this is just what the web app will serve at one time.

  `"batch_size": 10,
  "history_size": 10,`

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -0,0 +1,172 @@
...
+    Stream is expected to contain `text` and may contain image_path` and a `meta`
+    dictionary. Returns a generator of the stream.
+    """
+
+    nlp = spacy.blank("en")  # use blank spaCy model for tokenization
+
+    # ensure image prefix does not have a trailing slash
+    if image_prefix is None:
+        image_prefix = ""
+    image_prefix = image_prefix.rstrip("/")
```
laurejt, line 55: This should be fine as is, but the trailing slash could be a different character depending on the file system.

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -0,0 +1,172 @@
...
+            doc = nlp(task["text"])
+            task["tokens"] = [
+                {
+                    "text": token.text,
+                    "start": token.idx,
+                    "end": token.idx + len(token.text),
+                    "id": i,
+                }
+                for i, token in enumerate(doc)
+            ]
```
laurejt, line 68: This list comprehension is difficult to read, consider converting into a for loop for readability

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -0,0 +1,172 @@
...
+created with page-level text annotation in mind, and support annotating
+text with a reference image displayed beside the text (`annotate_page_text`),
+or annotating both text and image side by side (`annotate_text_and_image`).
+
+Referenced images must be served out independently for display.
+
+Example use:
+```
+prodigy annotate_page_text poetry_spans poetry_pages.jsonl --label POETRY,PROSODY -F ../corppa/poetry_detection/annotation/recipe.py --image-prefix http://localhost:8000/
+prodigy annotate_text_and_image poetry_text_image poetry_pages.jsonl --label POETRY,PROSODY -F ../corppa/poetry_detection/annotation/recipe.py --image-prefix http://localhost:8000/
```
laurejt, line 12: Consider removing the second label for these examples, since these tasks will only have one in the main use case

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -0,0 +1,172 @@
...
+    # copy the common config options and add blocks and labels
+    config = PRODIGY_COMMON_CONFIG.copy()
+    config.update(
+        {
+            "blocks": blocks,
+            "labels": label_list,
+            "image_manual_spans_key": "image_spans",
+            # limit image selection to rectangle only, no polygon or freehand
+            "image_manual_modes": ["rect"],
+        }
```
laurejt, line 112: We might also want to set the following config options:
- `"darken_image": 0.25` This will darken the area within the bound box by 25% (we might want a different value)
- `"image_manual_show_labels": false` Don't show the labels by default, since this task should only have one by default.

src/corppa/poetry_detection/annotation/add_metadata.py
```diff
@@ -0,0 +1,102 @@
...
+        # TODO: how do we want to handle image path ?
+        # replace .tif with .jpg in image_path (quick fix for the image path issue)
+        page["image_path"] = page["image_path"].replace(".TIF", ".jpg")
+
+        yield page
+
+
+def main():
+    """Add work metadata to pages for display in Prodigy"""
+    # Available as `corppa-??` when this package is installed with pip."""
```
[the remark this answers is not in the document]
rlskoeser: oh, yeah, that's when I thought this should be dropped in as an existing script before I knew how much overlap there were between our various silos of scripts. This is in corppa but I haven't added a pyproject config to make it available as a command.

src/corppa/poetry_detection/annotation/add_metadata.py
```diff
@@ -0,0 +1,102 @@
...
+        "output",
+        help="Filename where the updated corpus should be saved",
+        type=pathlib.Path,
+    )
+    parser.add_argument(
+        "--progress",
+        help="Show progress",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+    )
```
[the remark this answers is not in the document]
rlskoeser: `--no-progress` ; using `--help` will show this option

src/corppa/poetry_detection/annotation/recipe.py
rlskoeser: these are urls rather than paths, so that prodigy can reference the image we're serving out

laurejt: Oh, oops. I hadn't realized that these were urls rather than file paths. That said, I'd suggest changing the variable name `image_path` to `image_url` so it's clear it's a URL and not a filepath.

src/corppa/poetry_detection/annotation/prodigy.css
```diff
@@ -0,0 +1,43 @@
...
+
+/* labels for first block (image) are in their own div */
+.prodigy-title-wrapper {
+  grid-area: header;
+}
+/* display metadata at top */
+.prodigy-meta {
+	grid-area: meta;
+	margin: 0 auto;
+}
```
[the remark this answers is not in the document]
rlskoeser: I'm going to leave this for now but note that if we change the text size we'll likely need to adjust the grid layout as well.

src/corppa/poetry_detection/annotation/recipe.py
laurejt: ...but why do we need a label at all?

PR #76, opened 2024-08-20
Script for copying & transforming images

laurejt:


rlskoeser (approved):
Slick. This looks great for what we need now, and we can figure out how/if to revise when we do the next round.

PR #78, opened 2024-08-22
Revised OCR script and related package files

laurejt:


src/corppa/ocr/gvision_ocr.py
```diff
@@ -7,60 +7,168 @@
...
+except ImportError:
+    print(
+        "Error: Python environment does not contain google-cloud-vision "
+        "package. Switch environments or install package and try again."
+    )
+    sys.exit(1)
+
+
+def image_relpath_generator(image_dir, ext_set, follow_symlinks=True):
+    for dirpath, dirs, files in image_dir.walk(follow_symlinks=follow_symlinks):
```
rlskoeser, line 30: please add comments briefly explaining what you told me verbally! so helpful and valuable to document  why you're using walk instead of glob (single instead of multiple for multiple extensions, symlink handling)

src/corppa/ocr/gvision_ocr.py
```diff
@@ -7,60 +7,168 @@
...
+        type=int,
+        default=0,
+    )
+    parser.add_argument(
+        "--ext",
+        help="Accepted file extension(s). Defaults: .TIF, .jpg",
+        nargs="*",
+        type=str,
+        action="extend",
+    )
```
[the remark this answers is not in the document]
rlskoeser: have a set of allowed images; command line arg is only if you want to restrict to a single

src/corppa/ocr/gvision_ocr.py
```diff
@@ -7,60 +7,168 @@
...
+
+    args = parser.parse_args()
+    # Workaround: Set default extensions if none are provided.
+    if args.ext is None:
+        args.ext = [".TIF", ".jpg"]
+
```
rlskoeser, line 150: solution from discussion! add a new optional argument to specify volume ids and use the helper utils you wrote (move from scripts until somewhere like `corppa.utils.paths` or `corppa.utils.pathutils`) to determine the correct path.  Then add and call a different function instead of the `get_image_ocr` that will run on everything.

src/corppa/ocr/gvision_ocr.py
```diff
@@ -7,60 +7,168 @@
...
+    ocr_count = 0
+    for image_relpath in image_relpath_generator(in_dir, set(ext_list)):
+        # Refresh progress bar
+        if show_progress:
+            progress_bar.refresh()
+        imagefile = in_dir.joinpath(image_relpath)
+        textfile = out_dir.joinpath(image_relpath).with_suffix(".txt")
+        jsonfile = out_dir.joinpath(image_relpath).with_suffix(".json")
+
+        # Request OCR if file does not exist
```
rlskoeser, line 66: starting here looks like roughly the block of logic to pull out into a separate function that you can call from either mode of running this script (specific volumes or not)

src/corppa/ocr/gvision_ocr.py
```diff
@@ -7,60 +7,168 @@
...
+        args.ext = [".TIF", ".jpg"]
+
+    # Validate arguments
+    if not args.input.is_dir():
+        print(f"Error: input directory {args.input} does not exist")
+        sys.exit(1)
+    # TODO: Is this too restrictive / unnecessary?
+    if not args.output.is_dir():
+        print(f"Error: output directory {args.output} does not exist")
+        sys.exit(1)
```
rlskoeser, line 158: might be nice to have an option to create it for you but I'd rather be warned than have it happily put things somewhere I didn't intend!

src/corppa/ocr/gvision_ocr.py
```diff
@@ -7,60 +7,168 @@
...
+        "package. Switch environments or install package and try again."
+    )
+    sys.exit(1)
+
+
+def image_relpath_generator(image_dir, ext_set, follow_symlinks=True):
+    for dirpath, dirs, files in image_dir.walk(follow_symlinks=follow_symlinks):
+        # Check the files in walked directory
+        for file in files:
+            ext = os.path.splitext(file)[1]
```
rlskoeser, line 33: I think there's probably a pathlib way to do this. [suffix](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix) maybe?

rlskoeser (commented):
Great stuff!

src/corppa/ocr/gvision_ocr.py
laurejt: Not in this case. `file` is a string, so I'd have to create an additional pathlib.Path object in order to use pathlib which seemed excessive.

src/corppa/ocr/gvision_ocr.py
laurejt: I think it depends on our expected use case, since this is the top-level directory. It seems important that it exists to avoid the mistake where the output directory is wrong for some reason so any previous OCR efforts cannot be found and will be fully duplicated in the "new" output directory.

src/corppa/ocr/gvision_ocr.py
laurejt: Oh, oops. I think we were thinking about how we would tackle this. Hopefully the latest revisions will address this sufficiently.

src/corppa/ocr/gvision_ocr.py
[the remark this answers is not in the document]
laurejt: Turns out there isn't much issue having multiple optional `nargs`. That said, if we want to support the option to provide multiple input directories we need to think through what the desired behavior is for the singular output directory.

src/corppa/ocr/gvision_ocr.py
```diff
@@ -1,66 +1,293 @@
...
+
+        # OCR images
+        report = ocr_images(
+            in_vol_dir,
+            out_vol_dir,
+            ext_list,
+            ocr_limit=current_ocr_limit,
+            show_progress=show_progress,
+        )
+
```
rlskoeser, line 191: Ah, I see - this is exactly equivalent to the way you were thinking of running it manually, you create the parallel output directories and then run the main `ocr_images` with the input and output directory for this volume. That's not how I'd imagined doing it, which is why I had that comment about potentially pulling out the ocr api chunk of code for reuse (we may still want to do that if it makes unit testing easier). 

It would be nice if we could make a generator of all the images from the specified volumes and have a single progress bar across all the pages form the volumes. Based on how you've structured this code, I think you could make a volume image generator that yields the paths for each volume and makes them relative to the base input directory (not sure yet how you would elegantly configure/pass the generator to the ocr image). I think it would be less code,  but not sure it's worth refactoring at this point unless it simplifies testing.

src/corppa/ocr/gvision_ocr.py
```diff
@@ -1,66 +1,293 @@
...
+            if current_ocr_limit == 0:
+                if show_progress:
+                    print("Hit OCR limit.", file=sys.stderr)
+                break
+
+    print(
+        f"---\nIn total, {total_ocr_count:,} images OCR'd & {total_skip_count:,} "
+        "images skipped.",
+        file=sys.stderr,
+    )
```
rlskoeser, line 207: Thanks for adding the reporting, this is great

test/test_utils/test_path_utils.py
```diff
@@ -0,0 +1,58 @@
+import pathlib
+import pytest
+
+from corppa.utils.path_utils import encode_htid, decode_htid
+from corppa.utils.path_utils import get_stub_dir, get_vol_dir
```
rlskoeser, line 5: preferred python style is to group all imports from the same module
```suggestion
from corppa.utils.path_utils import encode_htid, decode_htid, get_stub_dir, get_vol_dir
```

Also, I just discovered that our `ruff` pre-commit hook is not fixing imports for us. (In other projects we've used black and isort together, and I knew that ruff could cover both but didn't know we had to opt into it.) Sound ok to you if we turn this on?

test/test_utils/test_path_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def test_encode_htid():
```
rlskoeser, line 8: these unit tests look great - nice to have confirmed functionality and expected error cases

src/corppa/ocr/gvision_ocr.py
```diff
@@ -1,66 +1,293 @@
...
+        else:
+            bar_format = "{desc}: {n:,} images OCR'd | elapsed: {elapsed}, {rate_fmt}"
+            progress_bar = tqdm(
+                desc=desc, bar_format=bar_format, maxinterval=maxinterval
+            )
 
-        # if text file does not exist, request ocr
+    ocr_count = 0
+    skip_count = 0
+    for image_relpath in image_relpath_generator(in_dir, set(ext_list)):
```
rlskoeser, line 78: did you consider wrapping tqdm around the image generator? might simplify the code slightly

src/corppa/ocr/gvision_ocr.py
```diff
@@ -1,66 +1,293 @@
...
+
+# Attempt to import Google Cloud Vision Python Client
+try:
+    from google.cloud import vision
+except ImportError:
+    print(
+        "Error: Python environment does not contain google-cloud-vision "
+        "package. Switch environments or install package and try again."
+    )
+    sys.exit(1)
```
rlskoeser, line 24: we might need to move this but I think let's wait and figure it out when we look at unit testing

test/test_utils/test_path_utils.py
laurejt: Sounds fine with me, I had split these into two lines due to line length; but I think `ruff` is more lenient with line length (allows for a bit over 80)

src/corppa/ocr/gvision_ocr.py
laurejt: Honestly, I don't quite remember; I think this might be a holdover from when I there wasn't a standalone generator.

src/corppa/ocr/gvision_ocr.py
```diff
@@ -1,66 +1,293 @@
...
+    current_ocr_limit = ocr_limit
+    total_ocr_count = 0
+    total_skip_count = 0
+    for i, vol_id in enumerate(vol_ids):
+        vol_src = get_ppa_source(vol_id)
+        # Currently, this only supports Gale; skip other (i.e. HathiTrust) sources
+        if vol_src != "Gale":
+            print(
+                f"Warning: Skipping volume, since its source '{vol_src}' "
+                "is not yet supported.",
```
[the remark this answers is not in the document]
laurejt: Yes, this is an issue with the current HathiTrust directory structure on tigerdata. The issue is that right now the HathiTrust volume-level directory names include timestamps corresponding to the "version" timestamp.

In my mind this is a temporary issue, it'll work as soon as the volume-level directories correspond to the volumes "cleaned" id. One of the simplest fixes would be to move the version to a metadata file within the volume-level directory. That said, I think there's a larger question about how we want to handle the HathiTrust images since they aren't static like Gale.

test/test_utils/test_path_utils.py
```diff
@@ -0,0 +1,80 @@
...
+            NotImplementedError, match="HathiTrust volume directory conventions TBD"
+        ):
+            get_vol_dir(vol_id)
+        mock_get_stub_dir.assert_not_called()
+    else:
+        with pytest.raises(ValueError, match=f"Unknown source '{source}'"):
+            get_vol_dir(vol_id)
+        mock_get_stub_dir.assert_not_called()
+
+    mock_get_ppa_source.assert_called_with(vol_id)
```
rlskoeser, line 80: I love that you're putting into practice the stuff we talked about (mocks and parameterized tests). I'm not sure that parametrize is getting you so much in this case, since you can't parametrize the expected behavior.  It would probably be cleaner and more readable to have separate test cases than use the if statement.

I was puzzling over what to suggest here - you could possible add more parameters for call args (`None` or `False` for no call? or parameter for call count?) and an optional exception (`None` for the first case?) and then do a simplified version of what you have here now, but the errors and exception matching would be hard to do without making the parameters more complicated.

rlskoeser (approved):
The cleanup and tests you've done so far look good. I think we should get this merged so you can use continue use it without complicating your new feature development, and then we can do a separate branch for additional unit tests, and pair together to figure out the tricky bits around the google cloud vision library.

My idea about a generator for volume images isn't necessary - I think it would be easier for me to do a first pass at implementing it than try to explain it, since I haven't quite worked out how to integrate it. We could pair on an experimental version if you want to think through it together, but what you've implemented here works well and gives you the output you want when you're running specific volumes.

test/test_utils/test_path_utils.py
laurejt: Good point, I had hoped for there to be more reuse than there ended up being. I split this into three separate unit tests.

PR #80, opened 2024-09-04
Feature/build poem corpus

laurejt:
Things to note:
- This is a work in progress (sorry for the lack of comments, I know I need to add more)
- `chadwyck.yml` defines the conda environment that the scripts rely on


src/corppa/poetry_detection/chadwyck_healey/get_tag_stats.py
```diff
@@ -0,0 +1,179 @@
...
+                "tag": output["tag"],
+                "attr": output["attr"],
+                "attr_val": output["attr_val"],
+                "poems": ", ".join(output["poems"]),
+            }
+            writer.writerow(row)
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Build Chadwyck Healey (sub)corpus.")
```
rlskoeser, line 125: Needs a new description

rlskoeser (approved):
This seems fine for what we need at this point. It would help a lot to have some brief documentation or comments about the different methods, and some example outputs.   I've added comments and suggestions but there's nothing that really needs to be changed now.  

I'm curious about your rational for the conda env and keeping the dependencies separate, but again seems fine for what this is.

It's a little hard to follow the bs4 text extraction (probably similar to how you felt about my eulxml/lxml code). I'm surprised there isn't a simpler bs4ish way to get the text while omitting tags you don't want to include, it seems like something beautifulsoup would provide.

If we wanted to generalize later it looks like there's enough common functionality between the two scripts to create a shared utilities module that both scripts could use.

PR #93, opened 2024-09-25
Updated recipe

laurejt:
(1) allow character-level highlighting
 (2) drop the ignore button
 (3) update task routing to limit 2 annotations per task.

### DIFF
--- a/src/corppa/poetry_detection/annotation/recipe.py
+++ b/src/corppa/poetry_detection/annotation/recipe.py
@@ -24,6 +24,7 @@ CURRENT_DIR = Path(__file__).parent.absolute()
 
 #: common prodigy configurations for both recipes; copy and add blocks and labels
 PRODIGY_COMMON_CONFIG = {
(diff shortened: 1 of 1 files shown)


rlskoeser (approved):

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -38,6 +39,9 @@
         "hide_true_newline_tokens": False,
     },
     "global_css_dir": CURRENT_DIR,
+    # Task annotation flags, these should be move elsewhere
+    "feed_overlap": False,
+    "annotations_per_tasks": 2,
```
rlskoeser, line 44: @laurejt should this be `annotations_per_task` rather than task**s** ?

laurejt: Yes

rlskoeser: Sorry I didn't catch that yesterday! Reviewed too quickly, I guess

PR #98, opened 2024-10-01
Feature/ocr testing

laurejt:


src/corppa/ocr/gvision_ocr.py
```diff
@@ -16,47 +16,99 @@
...
+    out_json paths respectively.
+    """
+    # Check that Google Cloud Vision Python Client was successfully imported
+    if google_vision is None:
+        print(
+            "Error: Python environment does not contain google-cloud-vision "
+            "package. Switch environments or install package and try again.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
rlskoeser, line 67: It looks like this check was copied when you refactored out the method but no longer needed. It will already be checked before this function is called, since the client is now passed in. 
```suggestion
```

rlskoeser: Although I do see that we call other methods on `google_vision` within this method, so it may be reasonable to check. If you think it's a concern (or removing it messes up any unit test assumptions), then you could just refactor the duplicated check to a method that's called in both places.

rlskoeser (approved):
Refactor and unit tests look good. One comment about the duplicated google vision check that at first glance seemed unneeded, it's fine if you decide how you want to handle this. If you keep it, would be nice to have a single method for the check and add a rief comment in the `ocr_image_via_gvision` method why it's needed.

src/corppa/ocr/gvision_ocr.py
laurejt: Sure I can refactor it into a separate method, and have it invoked in both places. I think it's important to ensure the check is done in both places, primarily for any future refactoring since `ocr_image_via_gvision` relies directly on `google_vision` and not just the passed in client.

That said, maybe the `ocr_image_via_gvision` thing should gain state and contain the client itself.... but that might be out-of-scope for the current effort.

PR #100, opened 2024-10-03
Feature/round two prep

laurejt:
This includes my data prep scripts and additional recipe changes for round two.

### DIFF
--- a/src/corppa/utils/generate_page_set.py
+++ b/src/corppa/utils/generate_page_set.py
@@ -32,7 +32,9 @@ def get_pages(in_csv, k):
     with poetry are included and the remaining pages are selected randomly from
     the remaining pages under consideration.
 
-    Note: In some cases the generated page set may not match k.
+    Notes:
+        * In some cases the generated page set may not match k.
+        * This is not compatible with PPA works with non-sequential page ranges.
     """
     # Load page data
     page_pool = {}
@@ -62,6 +64,7 @@ def get_pages(in_csv, k):
         )
 
     # Select remaining pages randomly
+    # TODO: Revisit to simply page selcection logic
     while page_counter < k:
         # Select work
         work_id = random.choice(list(page_pool.keys()))
--- a/src/corppa/utils/generate_page_set.py
+++ b/src/corppa/utils/generate_page_set.py
@@ -1,0 +1,127 @@
+"""
+Utility for generating a PPA page set.
+
+This method takes three inputs: (1) an input csv, (2) an output csv, and
+(3) the size of the page set.
+
+The input CSV file must have the following fields:
+    * work_id: PPA work id
+    * page_start: Starting index for page range being considered for this work
+    * page_end: Ending index for page range being considered for this work
+    * poery_pages: Comma separated list of page numbers containing poetry
+
+The pages are selected as follows:
+    * First, all pages with poetry are selected
+    * Then, all remaining pages are chosen randomly (proportionately by work)
+
+The resulting output CSV file has the following fields:
+    * work_id: PPA work id
+    * page_num: Digital page number
+"""
+
+import argparse
+import csv
+import random
+import sys
+from pathlib import Path
+
+
+def get_pages(in_csv, k):
+    """
+    Using the input CSV, generate a page set with k pages such that all pages
+    with poetry are included and the remaining pages are selected randomly from
+    the remaining pages under consideration.
+
+    Note: In some cases the generated page set may not match k.
+    """
+    # Load page data
+    page_pool = {}
+    poetry_pages = {}
+    page_counter = 0
+    with open(in_csv, newline="") as csvfile:
+        reader = csv.DictReader(csvfile)
+        for row in reader:
+            work_id = row["work_id"]
+            if work_id not in page_pool:
+                page_pool[work_id] = {}
+            start_idx = int(row["page_start"])
+            end_idx = int(row["page_end"]) + 1
+            # Gather full page range
+            for i in range(start_idx, end_idx):
+                page_pool[work_id][i] = {"work_id": work_id, "page_num": i}
+            # Yield pages with poetry
+            for pg_id in row["poetry_pages"].split(","):
+                yield page_pool[work_id].pop(int(pg_id))
+                page_counter += 1
+
+    # Print warning if more poetry pages than k
+    if page_counter >= k:
+        print(
+            f"Warning: Too many pages with poetry (k = {k} <= {page_counter})",
+            file=sys.stderr,
+        )
+
+    # Select remaining pages randomly
+    while page_counter < k:
+        # Select work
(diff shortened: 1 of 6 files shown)


src/corppa/utils/path_utils.py
```diff
@@ -87,3 +87,30 @@ def get_vol_dir(vol_id):
...
+    Extract volume id from PPA work id
+
+    * For full works, volume ids and work ids are the same.
+    * For excerpts, the work id is composed of the prefix followed by "-p" and
+      the starting page of the excerpt.
+    """
+    return work_id.rsplit("-p", 1)[0]
+
+
+def get_image_relpath(work_id, page_num):
```
rlskoeser, line 103: thanks for moving this method here, this seems like a good location

src/corppa/utils/path_utils.py
```diff
@@ -87,3 +87,30 @@ def get_vol_dir(vol_id):
...
+    """
+    Get the (relative) image path for specified PPA work page
+    """
+    vol_id = get_volume_id(work_id)
+    vol_dir = get_vol_dir(vol_id)
+    source = get_ppa_source(vol_id)
+    if source == "Gale":
+        image_name = f"{vol_id}_{page_num:04d}0.TIF"
```
rlskoeser, line 111: TIF is the original extension, do we need a way to customize extension when we call this method or should it be handled somewhere downstream ?

src/corppa/utils/generate_page_set.py
```diff
@@ -0,0 +1,127 @@
...
+Utility for generating a PPA page set.
+
+This method takes three inputs: (1) an input csv, (2) an output csv, and
+(3) the size of the page set.
+
+The input CSV file must have the following fields:
+    * work_id: PPA work id
+    * page_start: Starting index for page range being considered for this work
+    * page_end: Ending index for page range being considered for this work
+    * poery_pages: Comma separated list of page numbers containing poetry
```
rlskoeser, line 11: I don't know if my comment belongs here or on the google doc you shared with your steps for constructing the dataset, but how did you construct the input CSV?

src/corppa/utils/generate_page_set.py
```diff
@@ -0,0 +1,127 @@
...
+    poetry_pages = {}
+    page_counter = 0
+    with open(in_csv, newline="") as csvfile:
+        reader = csv.DictReader(csvfile)
+        for row in reader:
+            work_id = row["work_id"]
+            if work_id not in page_pool:
+                page_pool[work_id] = {}
+            start_idx = int(row["page_start"])
+            end_idx = int(row["page_end"]) + 1
```
rlskoeser, line 48: FWIW, some PPA supports non-sequential page ranges. It probably doesn't matter in this case, but would be good in future to use the same `intspan` python library for consistency and simplicity.

src/corppa/utils/generate_page_set.py
```diff
@@ -0,0 +1,127 @@
...
+        # Select work
+        work_id = random.choice(list(page_pool.keys()))
+        # Select page
+        try:
+            pg_id = random.choice(list(page_pool[work_id].keys()))
+        except IndexError:
+            # Encountered empty list, remove work entry and continue
+            del page_pool[work_id]
+            continue
+        yield page_pool[work_id].pop(pg_id)
```
rlskoeser, line 75: Seems like you could simplify this by using `random.choices` or `random.sample`

src/corppa/utils/generate_page_set.py
```diff
@@ -0,0 +1,127 @@
...
+    * page_end: Ending index for page range being considered for this work
+    * poery_pages: Comma separated list of page numbers containing poetry
+
+The pages are selected as follows:
+    * First, all pages with poetry are selected
+    * Then, all remaining pages are chosen randomly (proportionately by work)
+
+The resulting output CSV file has the following fields:
+    * work_id: PPA work id
+    * page_num: Digital page number
```
rlskoeser, line 19: oh, I see, nice - this is the exact format we now support in the filter script

rlskoeser (approved):
This all looks fine. The unit tests for the path utils are good.  

I made some comments about possible improvements but none of those changes need to be made before merging.

I have some concerns about the continuing proliferation of scripts, but much prefer them to be in the repo than not. 

The selection of random pages by volume is a clever solution. Right now it's generating an output file that can be used as input to the filter script. I'm wondering what it might look like to connect the methods so it didn't have to be separate calls. Not proposing to add more functionality to the already complicated filter script, more thinking about whether we could make the internal logic reusable elsewhere.

Thanks for making notes about your steps to generate the dataset for this round. Once the dataset and revised recipe are deployed, it would be great for us to take some time to make sure all the details are clearly documented and think about what we want to improve for the next round. It still seems pretty hodgepodge at this point.

src/corppa/utils/generate_page_set.py
laurejt: I think this comes down to how we want to randomly select the additional pages. It seems like it'd be nice to do a weight sample so that there's a uniform chance of drawing a page from any of the `k` volumes. I know how to do this in `numpy`, but I wasn't sure it didn't seem like there was a nice way to do this with `random`

src/corppa/utils/generate_page_set.py
laurejt: I'll make a comment noting this.

src/corppa/utils/generate_page_set.py
laurejt: I'll make a comment in the google doc, but it was a relatively simple generation from the poem-focused test set spreadsheet.

src/corppa/utils/path_utils.py
laurejt: For now, I made the assumption that this is handled downstream. For example, the `add_image_relpaths` script takes an optional extension as input to replace the preexisting one.

laurejt:
> I have some concerns about the continuing proliferation of scripts, but much prefer them to be in the repo than not.

When we have time, we could consider combining many of these utilities into an "app"
 of sorts, like how Mallet and the PPA django app work.

src/corppa/utils/generate_page_set.py
laurejt: I will note that this seems like a potentially large refactor. I was going off of the poem-focused PPA test set spreadsheet, which had a single range for pages to include; but we can revisit this in the future.

rlskoeser: I don't think it's too big of a refactor, you just parse the range with intspan and let it generate the list of numbers in range instead of doing it based on start and end - but a comment is perfect, and we can assess whether we need it for the next round, it may still not matter (or we may be able to avoid it).

rlskoeser: If you are making an issue or document with notes and tag me I will add a comment there

PR #102, opened 2024-10-07
Fix exit code values to 1 and print error messages to stdout

laurejt:


rlskoeser (approved):
Good refactor, thank you for cleaning this up

Issue #75, opened 2024-08-19
load adjudication recipe on local environment


mnaydan:
Recipe needs to allow for character-level text span annotations, which the original annotators for round 1 couldn't do because they did tokens. 

If possible, we should use Prodigy's built-in review recipe: [https://prodi.gy/docs/recipes#review](https://prodi.gy/docs/recipes#review)

laurejt:
**For testing of the Adjudication recipe:**
- [ ] Displays annotations with session-name prefix (numerical suffix for "duplicate" sessions)
- [ ] Unique colors for annotators
- [ ] Character-based highlighting
- [ ] Progress bar reflects completed annotations 
- [ ] Cannot submit flagged task
- [ ] Cannot submit task with overlapping spans

PR #106, opened 2024-10-23
Adjudication Recipe

laurejt:
ref: #75 

* Added recipe for adjudicating test span annotations
* Updated other recipes to use Prodigy API and optionally fetch media
* Added unit testing for methods used by recipes

### DIFF
--- a/src/corppa/poetry_detection/annotation/recipe.py
+++ b/src/corppa/poetry_detection/annotation/recipe.py
@@ -317,7 +317,10 @@ def has_span_overlap(example: TaskType, strip_label_pfx: bool = True) -> bool:
             label = remove_label_prefix(label)
         # Represent span's coverage as an intspan
         # Note: intspans include the ending index unlike span character ranges
-        #       which follows Python range behavior
+        #       which follows Python range behavior.
+        # Ex. The set {1,2,3} is represented as follows:
+        #   * intspan("1-3") <--> interval [1,3]
+        #   * span: {"start": 1, "end": 4} <--> interval [1, 4)
         span_coverage = intspan.from_range(span["start"], span["end"] - 1)
         # Check if span's coverage is disjoint from label's existing coverage.
         if label not in label_coverage:


src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -41,129 +57,373 @@
...
+
     # copy the common config options and add blocks and labels
-    config = PRODIGY_COMMON_CONFIG.copy()
+    config = deepcopy(PRODIGY_COMMON_CONFIG)
     config.update(
         {
             "blocks": blocks,
-            "labels": label_list,
             "ner_manual_highlight_chars": True,
+            "global_css_dir": CURRENT_DIR,
```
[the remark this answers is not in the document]
laurejt: Good catch, I wasn't originally copying the full common config.

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -41,129 +57,373 @@
...
+    """
+    Set the image field for each example in the stream
+
+    Calls: `add_image`
+    """
+    for example in examples:
+        yield add_image(example, image_prefix=image_prefix)
+
+
+def remove_images(
```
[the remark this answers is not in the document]
laurejt: If I said something about removing URIs would that work? I'm removing the fetched media, and turning them back into what they were previously.

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -41,129 +57,373 @@
...
+        # TODO: Make sure that no unmerged version content is preserved.
+        review_example = deepcopy(versions[-1])
+        # Merge spans
+        merged_spans = []
+        session_counts = {}
+        sessions = []
+
+        for version in versions:
+            session_id = version[SESSION_ID_ATTR]
+            # Assume: session name does not contain -
```
rlskoeser, line 285:
```suggestion
            # Assume: session name does not contain -
            # full session name includes the dataset id; split to get the session name without dataset id
```

src/corppa/poetry_detection/annotation/recipe.py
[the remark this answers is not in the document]
rlskoeser: Maybe just rename to `remove_image_data`? Your docstring here is clear and it became more meaningful once I saw it used in context multiple times.

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -41,129 +57,373 @@
...
+                session_counts[session_id] += 1
+            sessions.append(session_name)
+            if "spans" not in version:
+                # Not sure when an annotated example would be missing a spans field
+                continue
+            for span in version["spans"]:
+                new_span = span.copy()
+                span_label = span["label"]
+                new_span["label"] = f"{session_name}: {span_label}"
+                merged_spans.append(new_span)
```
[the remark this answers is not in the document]
laurejt: Well, that helps answer how much clarity the type annotations provide. `versions` is a list of annotated tasks (`List[TaskType]`) to be merged. These tasks are grouped by input hash (this grouping occurs in `get_review_stream`. So, `version` is an annotated example (`TaskType`).

This code is making a single "merged" example for review and does it in the following way:
1. Copy the contents from one of the version
2. Then for each version:
  a. Determine the session name by removing the dataset prefix and adding a numerical suffix if a session has multiple annotated examples (hopefully only a headache for round 1)
  b. Copy its `spans` (but with modified label fields) to `merged_spans` list.
2. Set "merged" example `spans` to `merged_spans`
3. Add "sessions" field containing list of session names to "merged" example

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -234,12 +243,106 @@ def annotate_page_text(
...
+        components["before_db"] = lambda x: remove_image_data(
+            x, image_prefix=image_prefix
+        )
 
     return components
 
 
+def get_session_name(example: TaskType, suffix: Optional[str] = None) -> str:
+    """
+    Extract session name from task examplke. Session ids have the following form:
```
rlskoeser, line 257: minor typo
```suggestion
    Extract session name from task example. Session ids have the following form:
```

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -234,12 +243,106 @@ def annotate_page_text(
...
+    if "spans" not in example:
+        return False
+    label_coverage = {}  # label (str) --> coverage (intspan)
+    for span in example["spans"]:
+        # Get label
+        label = span["label"]
+        if strip_label_pfx:
+            label = remove_label_prefix(label)
+        # Represent span's coverage as an intspan
+        # Note: intspans are closed intervals, while span character ranges are half-open
```
rlskoeser, line 319: what does "half-open" mean here?

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -234,12 +243,106 @@ def annotate_page_text(
...
+        span_coverage = intspan.from_range(span["start"], span["end"] - 1)
+        # Check if label's intspan is disjoint from
```
rlskoeser, line 321: is this comment in complete and/or in the wrong place?

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -234,12 +243,106 @@ def annotate_page_text(
...
+        if label not in label_coverage:
+            label_coverage[label] = span_coverage
+        else:
+            if span_coverage.isdisjoint(label_coverage[label]):
```
rlskoeser, line 325: very cool that `intspan` has a method for this!  (maybe some of the other functionality also useful for working with these spans)

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -234,12 +243,106 @@ def annotate_page_text(
...
+def validate_review_answer(example: TaskType):
+    """
+    Validate review answer:
+    * Example not flagged (i.e. flagged field not set)
+    * Text spans (by label) must not overlap
+    """
+    if example.get("flagged") is True:
+        raise ValueError("Currently flagged, unflag to submit.")
+    if has_span_overlap(example):
+        raise ValueError("Overlapping spans with the same label detected!")
```
rlskoeser, line 343: I think I understand the logic of the span overlap check but I'm missing why it's needed - is this something you and Mary agreed on? validation/adjudication requires cleaning up existing annotations so there is no overlap and one definitive set of spans?

rlskoeser (approved):
Looks good to me. There are a few places where comments might provide a little more context, but this seems to be in very good shape.

Nice job on refactoring and improving all the recipes and seriously beefing up the unit testing for this part of the code.

src/corppa/poetry_detection/annotation/recipe.py
laurejt: This is a customizable feature of Prodigy Recipes, to allow validation of answers *before* allowing an annotator to submit the annotation (see [documentation](https://prodi.gy/docs/custom-recipes#validate_answer)).

So, the `validate_review_answer` method is being used to validate the annotations being "submitted" in the interface (i.e., selecting the accept or reject buttons). When an error is raised, a pop-up window appears telling the annotator they cannot submit this "invalid" annotation. The error messages are also shown within this pop-up window.

src/corppa/poetry_detection/annotation/recipe.py
laurejt: Good catch, looks like the comment got cut-off. It should be
```suggestion
        # Check if span's coverage is disjoint from label's existing coverage.
```

src/corppa/poetry_detection/annotation/recipe.py
rlskoeser: ah, I see! makes sense - that's why the minus one on the end

src/corppa/poetry_detection/annotation/recipe.py
rlskoeser: I assumed it was something like that, although I didn't know about the UI integration - seems like a very powerful feature. My question was actually that I don't understand why overlapping spans  == invalid

test/test_poetry_detection/test_annotation/test_recipe.py
```diff
@@ -0,0 +1,411 @@
...
+sys.modules["prodigy.core"] = MagicMock()
+sys.modules["prodigy.errors"] = MagicMock()
+sys.modules["prodigy.types"] = MagicMock()
+sys.modules["prodigy.util"] = MagicMock()
+## Mock for set_hashes
+mock_prodigy = MagicMock()
+sys.modules["prodigy"] = mock_prodigy
+## Mock for fetch_media
+mock_prodigy_preprocess = MagicMock()
+sys.modules["prodigy.components.preprocess"] = mock_prodigy_preprocess
```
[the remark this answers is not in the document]
laurejt: Done!

src/corppa/poetry_detection/annotation/recipe.py
laurejt: So, this is to prevent duplicate and/or conflicting spans to exist when adjudicating. For our task, spans of the same label should not overlap (a passage can only be an excerpt of a single poem).

laurejt: This is primarily a validation check for the adjudicator, that they've culled enough of the existing annotations before submission.

src/corppa/poetry_detection/annotation/recipe.py
```diff
@@ -234,12 +243,106 @@ def annotate_page_text(
...
+                # Overlap found
+                return True
+    return False
+
+
+def validate_review_answer(example: TaskType):
+    """
+    Validate review answer:
+    * Example not flagged (i.e. flagged field not set)
+    * Text spans (by label) must not overlap
```
rlskoeser, line 338: Your explanation in the comments was really clear & helpful - it would be good to include a brief version of it in the comment here. 

```suggestion
    * Text spans with the same label must not overlap (no duplicates or conflicting spans)
```

laurejt: How's the updated documentation look?

src/corppa/poetry_detection/annotation/recipe.py
laurejt: Is the updated comment helpful?

src/corppa/poetry_detection/annotation/recipe.py
rlskoeser: great! thank you for clarifying

src/corppa/poetry_detection/annotation/recipe.py
rlskoeser: that notation isn't familiar enough to me to be that helpful here - can we just say that intspan includes the last character but span characters do not?

laurejt: fair enough, this is standard interval notation; but I can rephrase it. since these aren't actually characters but integer sets; i'll just say something about how intspan's range includes the ending index (as opposed to python indexing)

laurejt: How about now?

rlskoeser: I figured it was standard notation, but it's not something I use frequently. That's a helpful clarification, to differentiate from python indexing. Probably good to include both standard interval notation and your note about ending index.

Sorry to belabor this; it does seem important to be clear on since I expect we will be working with these spans in other contexts and I want to make sure I understand.

For my own understanding on your note about characters vs integer sets - it's integers that reference character based on indexes, right? I see now that I was sloppy in my language, was thinking of what the numbers represent rather than the numbers themselves

rlskoeser: your last revision is fine with me if you're happy with it!

laurejt: Thanks for the good conversation, I'll tweak the comment a bit more but it's definitely worth documenting.

laurejt: How's the added example look?

rlskoeser: this looks great! thank you

PR #107, opened 2024-10-29
Script to collate ocr text files by directory into one json file per directory (for Gale local OCR)

rlskoeser:
Updates are related to https://github.com/Princeton-CDH/ppa-django/issues/673

changes in this PR:

- refactored & updated `image_relpath_generator` originally written for the gogle vision ocr script as follows:
   -  renamed it (since there's nothing image specific in the logic)
   - moved it into our path utils for reuse
   - added an option to group files by directory
- new script that I called `collate_text`, which finds directories with text files and compiles them into json files in a parallel structure in a specified output directory (borrows heavily from the google visions script approach)


test/test_utils/test_path_utils.py
```diff
@@ -125,3 +127,141 @@ def test_get_image_relpath(mock_get_ppa_source, mock_get_vol_dir, mock_get_volum
...
+    # Do not follow symbolic links
+    paths = list(find_relative_paths(dir_a, [".jpg"], follow_symlinks=False))
+    assert {jpg_a, jpg_b} == set(paths)
+
+
+def test_find_relative_paths_group_by_dir(tmp_path):
+    ocr_dir = tmp_path / "ocr"
+    ocr_dir.mkdir()
+    vol1_dir = ocr_dir / "vol1"
+    vol2_dir = ocr_dir / "vol1"
```
laurejt, line 230: Are these directories supposed to be distinct? Currently, they are the same directory.
```suggestion
    vol2_dir = ocr_dir / "vol2"
```

src/corppa/ocr/collate_txt.py
```diff
@@ -0,0 +1,118 @@
...
+    # NOTE: this logic is currently specific to Gale/ECCO files,
+    # which look like CW0112029406_00180.txt
+
+    # split the file base/stem name by _ and take the last part
+    pagenum = filename.stem.split("_")[-1]
+    # return the number as a string; strip extra trailing zero
+    return pagenum[:-1]  # strip trailing zero
+
+
+def collate_txt(
```
laurejt, line 27: `collate_txt` missing documentation

src/corppa/ocr/collate_txt.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    for ocr_dir, files in tqdm(
+        find_relative_paths(input_dir, [".txt"], group_by_dir=True),
+        desc="Directories with text files",
+        disable=not show_progress,
+    ):
+        # output will be a json file based on name of the directory containing text files,
+        # with parallel directory structure to the source
+        output_file = (output_dir / ocr_dir.parent / ocr_dir.name).with_suffix(".json")
```
laurejt, line 48: This was confusing for parsing what is the directory vs. filename. Here's an alternative:

```suggestion
        output_file = output_dir / ocr_dir.parent / f"{ocr_dir.name}.json"
```

src/corppa/ocr/collate_txt.py
```diff
@@ -0,0 +1,118 @@
...
+            continue
+
+        directories += 1
+        txt_files += len(files)
+        status.set_postfix_str(f" {ocr_dir.stem}: {len(files)} txt files")
+
+        # combine text contents into a dictionary keyed on page number
+        txt_data = {}
+        for filename in files:
+            with (input_dir / filename).open() as txtfile:
```
laurejt, line 61: More defensive coding (I'm sure I've written code not guarding for this too)
```suggestion
            with (input_dir / filename).open(encoding="utf-8") as txtfile:
```

src/corppa/ocr/collate_txt.py
```diff
@@ -0,0 +1,118 @@
...
+                txt_data[page_number(filename)] = txtfile.read()
+
+        # ensure the parent directory exists
+        output_file.parent.mkdir(exist_ok=True)
+        # save out text content as json
+        with output_file.open("w") as outfile:
```
laurejt, line 67: More defensive coding (I'm sure I've written code not guarding for this too)
```suggestion
        with output_file.open("w", encoding="utf-8") as outfile:
```

laurejt (commented):
Is it worth adding a way to collate a specific set of volumes?

src/corppa/ocr/collate_txt.py
rlskoeser: I had to look this up, it hasn't been my habit to worry about this - I see that it's still not the default despite a fair amount of debate about that. Am I right that it primarily affects code running on Windows ?

laurejt: Probably? I mostly worry about servers. On an old cluster there were a few machines that had latin-1 as their default (we found out in the worst way)

laurejt: It definitely will impact Windows machines, but there probably could be others.

src/corppa/utils/path_utils.py
```diff
@@ -118,6 +118,20 @@ def get_image_relpath(work_id, page_num):
...
+    file for content from a single page (e.g., image or text). Returns the
+    page number as a string with leading zeros. (Note: logic is curently
+    specific to Gale/ECCO file naming conventions.)"""
+    # NOTE: this logic is currently specific to Gale/ECCO files,
+    # which look like CW0112029406_00180.txt
+
+    # split the file base/stem name by _ and take the last part
+    pagenum = filename.stem.split("_")[-1]
+    # return the number as a string; strip extra trailing zero
+    return pagenum[:-1]  # strip trailing zero
```
laurejt, line 132: Is it worth refactoring this now to take a `source` string as input similar to `get_vol_dir` for future source support?

laurejt (approved):

src/corppa/utils/path_utils.py
```diff
@@ -118,16 +118,18 @@ def get_image_relpath(work_id, page_num):
         raise ValueError(f"Unsupported source '{source}'")
 
 
-def page_number(filename: pathlib.Path) -> str:
-    """Extract and return the page number from a :class:`pathlib.Path`
-    file for content from a single page (e.g., image or text). Returns the
-    page number as a string with leading zeros. (Note: logic is curently
+def get_page_number(filename: str) -> str:
```
laurejt, line 121: Was there a reason for switching from `pathlib.Path` to `str` as input? The equivalent of `os.path.splitext` is to just use `Path.stem` & `Path.ext`.

src/corppa/utils/path_utils.py
laurejt: Sorry for the red herring. It hadn't occurred to me that the volume identifier could be extracted from image filename, so the source could be inferred (rather than specified as a second input argument)

rlskoeser: We can revisit in future if that doesn't hold true for other collections.

src/corppa/utils/path_utils.py
rlskoeser: I misunderstood a previous comment, I thought you requested that I make that change. 

Just switched it back to pathlib. Shall we get this merged?

laurejt (approved):

src/corppa/utils/path_utils.py
laurejt: Looks good to me! Sorry for the confusion.

PR #109, opened 2024-10-31
Convenience function and script to output path to recipe file

rlskoeser:
@laurejt here's a quick version of what I was suggesting, to make it easier to get the path to the recipe file, since that's how prodigy requires we specify it.

I wasn't able to test this locally since I don't have prodigy installed, but I think it should work as I've written it into the examples and hopefully it's pretty easy for you to check.


rlskoeser:
Closing without merging since we decided not to go this direction.

PR #113, opened 2024-11-01
Customized stats command

laurejt:
ref: #112 

Primary: Additional command recipe for displaying progress specialized to our annotation task

Secondary: Renaming recipe files to better describe the kinds of Prodigy recipes they will contain.


src/corppa/poetry_detection/annotation/command_recipes.py
```diff
@@ -0,0 +1,87 @@
...
+        raise RecipeError(f"Can't find dataset '{dataset}' in database")
+    examples = DB.get_dataset_examples(dataset)
+    n_examples = len(examples)
+    msg.good(f"Loaded {n_examples} annotations from {dataset} dataset")
+
+    # Get stats
+    examples_by_page = Counter()
+    examples_by_session = defaultdict(list)
+    for ex in examples:
+        # Skip examples w/ answer or (page) id
```
rlskoeser, line 26: by "w/" here did you mean "w/o" / without?
```suggestion
        # Skip examples without answer or (page) id
```

src/corppa/poetry_detection/annotation/command_recipes.py
```diff
@@ -0,0 +1,87 @@
...
+        page_id = ex["id"]
+        examples_by_page[page_id] += 1
+        session_id = ex[SESSION_ID_ATTR]
+        examples_by_session[session_id].append(page_id)
+    # Get frequencies of page-level annotation counts
+    count_freqs = Counter()
+    total = 0
+    for count in examples_by_page.values():
+        count_freqs[count] += 1
+        total += count
```
rlskoeser, line 39: What you're doing here wasn't immediately obvious based on the variable names - am I understanding correctly that you're counting the number of examples/pages that have the same number of annotations, so you can report something like 100 examples have 1 annotation each, 50 have 2 annotation each, etc? 

You can let `Counter` do the aggregation for you by using `Counter(examples_by_page.values())` .

I think it would be more readable to tally this way:
```suggestion
    # Get frequencies of page-level annotation counts
    count_freqs = Counter(examples_by_page.values())
    total = Sum(examples_by_page.values())
```

src/corppa/poetry_detection/annotation/command_recipes.py
```diff
@@ -0,0 +1,87 @@
...
+        title="Overall Annotation Progress",
+        header=header,
+        aligns=aligns,
+        divider=True,
+    )
+
+    # Build session table
+    data = []
+    total = 0
+    for session, pages in sorted(examples_by_session.items()):
```
rlskoeser, line 61: Is this sorting so you display sessions in alpha order?

src/corppa/poetry_detection/annotation/command_recipes.py
```diff
@@ -0,0 +1,87 @@
...
+        "Total",
+    ]
+    aligns = ["l", "r", "r", "r"]
+    # info = {
+    #    "Session": "Session name",
+    #    "Count": "Completed annotations",
+    #    "Unique": "Unique annotations (distinct pages)",
+    #    "Total": "Total annotations collected",
+    # }
+    # msg.table(info, title="Legend")
```
rlskoeser, line 80: leftover comments to be cleaned up?

src/corppa/poetry_detection/annotation/command_recipes.py
```diff
@@ -0,0 +1,87 @@
...
+        divider=True,
+    )
+
+    # Build session table
+    data = []
+    total = 0
+    for session, pages in sorted(examples_by_session.items()):
+        count = len(pages)
+        unique = len(set(pages))
+        total += count
```
rlskoeser, line 64: Is the total cumulative here?  Maybe worth renaming the variable to clarify

src/corppa/poetry_detection/annotation/command_recipes.py
```diff
@@ -0,0 +1,87 @@
+from collections import Counter, defaultdict
```
rlskoeser, line 1: Would be helpful to add a docstring either at the top or with the recipe explaining how you run this and showing some sample output.

rlskoeser (approved):
This reporting recipe looks really great! Asked a few questions to clarify things and made a couple of small suggestions, but I think it's fine to merge whenever you're happy with it.

Question about structure, since you've moved things around a bit: what do you think about dropping the `poetry_detection` directory and just putting all of this in `corppa.annotation` ? There's nothing specific to poetry detection in the annotation recipes that I can think of.

src/corppa/poetry_detection/annotation/command_recipes.py
laurejt: Good catch!

src/corppa/poetry_detection/annotation/command_recipes.py
laurejt: Thanks, that's a good idea. Although the styling itself is a bit outside of my understanding (I believe it's rendered by an external library)

src/corppa/poetry_detection/annotation/command_recipes.py
laurejt: No, it's a design choice. Uncommented, this prints out the legend for this table, but it's fairly verbose.

src/corppa/poetry_detection/annotation/command_recipes.py
laurejt: I guess so? I don't remember the reason for this; it might be residual code since the other commands sorted things.

src/corppa/poetry_detection/annotation/command_recipes.py
laurejt: Fair enough, this is also residual code in terms of naming conventions for total.

laurejt: ...but yes, it's the total annotations collected as described [here](https://github.com/Princeton-CDH/ppa-nlp/blob/e35a19bf5781d7cf3c943f5d6a2bf15cd2f1a533/src/corppa/poetry_detection/annotation/command_recipes.py#L78)

laurejt: I can rename the variable to `cumulative_total` if it helps but that seems too long for the reporting output itself

src/corppa/poetry_detection/annotation/command_recipes.py
laurejt: These are the count frequencies, the frequency at which each page image has been annotated so far (like you described)

Good point on the `Counter` / `sum` usage. Not sure why I missed that.

src/corppa/poetry_detection/annotation/command_recipes.py
rlskoeser: That makes sense. 👍 to renaming as `cumulative_total`

src/corppa/poetry_detection/annotation/command_recipes.py
rlskoeser: ah, I see. Maybe add a comment about the comment, then, so someone else doesn't clean it up?

laurejt: I mean, I'm happy to remove it, but this was more of "do we want a legend" question that I had forgotten about.

rlskoeser:
As part of this refactor, I suggest moving out the tested part of the prodigy/annotation code into a some kind of shared utility methods file so that we can exclude the recipe code from code coverage. 

Here's the syntax for excluding files from codecov reporting: https://docs.codecov.com/docs/ignoring-paths

src/corppa/poetry_detection/annotation/command_recipes.py
laurejt: I'm going to switch the ordering, so its ordered by # of annotations in decreasing order

Issue #125, opened 2024-12-16
get initial passim pipeline into develop


mnaydan:
- [x] add readme of expected sequence of commands & output structure and additional comments in code
- [ ] make issue with things we want to improve later

laurejt:
Tabling creating additional issues until we January. Issue #84 covers the need to better intergrate the passim workflow.

PR #128, opened 2024-12-16
Initial passim pipeline

laurejt:
Ref: #125 

The passim pipeline involves the following steps:

1. Create passim-friendly input corpora using `create_passim_corpus.py`
2. Running "default" passim with `run_passim.sh`
3. Build page-level passim results with `build_passim_results.py`
4. Gather excerpt-level passim results using `get_passim_excerpts.py`


src/corppa/poetry_detection/passim/build_passim_results.py
```diff
@@ -0,0 +1,217 @@
...
+    parser.add_argument(
+        "--progress",
+        help="Show progress",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+    )
+    parser.add_argument(
+        "--include-excerpts",
+        help="Include text excerpts (original & aligned) for each match",
+        action="store_true",
```
rlskoeser, line 179: I vote for this being on by default. We could make it a boolean arg default true so you could do `--excerpts` / `--no-excerpts`

src/corppa/poetry_detection/passim/build_passim_results.py
```diff
@@ -0,0 +1,217 @@
+"""
+Gather passim results (i.e., identified matches) at the page-level from
+passim output files.
```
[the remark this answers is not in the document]
rlskoeser: figured out the ref corpus option by trial error from looking at the output, here's my latest syntax for running it:
```sh
python src/corppa/poetry_detection/passim/build_passim_results.py prelim-round-one/ppa_hathi_passim.jsonl prelim-round-one/passim_out/ prelim-round-one/passim_rebuilt3 --include-excerpts --ref-corpus prelim-round-one/ref_poems.jsonl
```

src/corppa/poetry_detection/passim/get_passim_excerpts.py
```diff
@@ -0,0 +1,82 @@
+"""
+Gather excerpts from passim page-level results.
```
rlskoeser, line 2: why make this separate from the other script for processing passim results?

src/corppa/poetry_detection/passim/get_passim_excerpts.py
```diff
@@ -0,0 +1,82 @@
...
+        ]
+        writer = csv.DictWriter(tsvfile, fieldnames=fieldnames, dialect="excel-tab")
+        writer.writeheader()
+        for match in get_passage_matches(input_file):
+            writer.writerow(match)
+
+
+def main():
+    """
+    Command-line access to build TSV file of passage-level passim matches.
```
rlskoeser, line 50: how well does tsv work for multi-line text?

if we keep this, in future I'd like to see an option for csv output also

src/corppa/poetry_detection/passim/get_passim_excerpts.py
```diff
@@ -0,0 +1,82 @@
...
+            "ref_id",
+            "ref_corpus",
+            "page_start",
+            "page_end",
+            "ref_start",
+            "ref_end",
+            "page_excerpt",
+            "ref_excerpt",
+            "aligned_page_excerpt",
+            "aligned_ref_excerpt",
```
rlskoeser, line 40: what are the differences= between the excerpts and the aligned excerpts? in most cases as I'm scanning they seem exactly the same - is one of them the version that includes the passim markings for extra/missing characters?

rlskoeser (approved):
Looked through the code; installed and ran through everything locally with some sample data. I think this just needs a little bit more documentation and example usage.

- ✅ Successfully ran create passim corpus script on a subset and full ppa pages
- ✅ Was able to install passim, pyspark, a supported java version, and ran the script
- ✅ ran the two scripts to process and augment passim results and am able to see preliminary results 

Everything works, I would just like to see a bit more documentation to make sure that I understand and we have it clear write up of the current state of these scripts and how they all fit together.  I think it would be best to add a brief readme to the passim directory with the expected sequence of commands to run with examples, rather than putting documentation in each individual script.

src/corppa/poetry_detection/passim/build_passim_results.py
```diff
@@ -0,0 +1,217 @@
...
+
+def get_span_annotation(alignment_record, include_excerpts=False):
+    span_annotation = {}
+    span_annotation["page_id"] = alignment_record["id2"]
+    span_annotation["ref_id"] = alignment_record["id"]
+    span_annotation["ref_corpus"] = alignment_record["corpus"]
+    span_annotation["page_start"] = alignment_record["begin2"]
+    span_annotation["page_end"] = alignment_record["end2"]
+    span_annotation["ref_start"] = alignment_record["begin"]
+    span_annotation["ref_end"] = alignment_record["end"]
```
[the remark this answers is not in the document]
laurejt: Sure, we can do that for the code but the paired ordering is much more readable for the output files themselves.

src/corppa/poetry_detection/passim/get_passim_excerpts.py
laurejt: The aligned excerpts are the character-level alignment output used by passim. They are never exactly the same "-" are used any time an insertion/removal happens in the alignment algorithm.

laurejt: Passim does not return the original page / reference text excerpts; only the "aligned" versions. That's why the external text corpora need to be provided because it's not included in the passim results files themselves.

src/corppa/poetry_detection/passim/get_passim_excerpts.py
laurejt: Perfectly fine. Tabs are better for all the command-line utilities, and arguably tabs are less common than commas in the texts we're processing.

I can switch to CSV for the sole reason that Google Docs has some broken exporting happen around TSVs (this is a short coming of that tool only)

src/corppa/poetry_detection/passim/get_passim_excerpts.py
laurejt: Because it was a secondary thing. We need page-level results full-stop, but it's nice to have the excerpts in a tsv/csv format so I can copy paste into a worksheet for manual stuff.

Fundamentally, some times we want to operate at the page-level and other times we want to operate at the excerpt-level, but not all pages even have excerpts. And we want to know about these cases (e.g., pages w/o any poems and pages for passim does not find any poesm).

laurejt: Another way to look at the outputs are for two separate things:

(1) Creating page-level evaluations.
(2) Identified poetry excerpts (for building the excerpt data set whose records fundamentally operate at the excerpt-level and not page-level).

src/corppa/poetry_detection/passim/build_passim_results.py
laurejt: I mean, we can just make it a required parameter; it's secondary because it serves not real purpose for evaluation. It's just nice for building the excerpt-level data.

src/corppa/poetry_detection/passim/create_passim_corpus.py
```diff
@@ -0,0 +1,177 @@
...
+    # Copy over all fields
+    if preserve_fields:
+        if id_field != "id" and "id" in record:
+            raise ValueError(f"Record has existing 'id' field")
+        out_record |= record
+    # Add required output
+    out_record["id"] = record[id_field]
+    out_record["corpus"] = corpus_name
+    # If text field is missing, treat as blank page
+    out_record["text"] = record.get("text", "")
```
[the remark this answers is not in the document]
laurejt: I mean, these files were adjudicated? It's not like it takes any space, I find it more troubling if a transformation drops records. I explicitly had to include these cases because the PPA text corpus has pages without "text" fields, but we included them in our annotation rounds.

src/corppa/poetry_detection/passim/create_passim_corpus.py
```diff
@@ -0,0 +1,177 @@
+"""
+Create a passim-friendly version of the input corpus.
```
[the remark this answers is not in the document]
laurejt: Sure, I though this was a preliminary review and not that this was supposed to be production ready. I can add an example.

src/corppa/poetry_detection/passim/run_passim.sh
```diff
@@ -0,0 +1,66 @@
...
+IN=$1
+OUT=$2
+
+
+if [ $# -ne 2 ]; then
+  echo "Usage: [input] [output]"
+  exit 1
+fi
+
+# Check for supported java version (via major version number)
```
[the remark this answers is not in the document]
laurejt: I mean, I did this because the error messages are *very* unclear. And this is a repeated issue for me because I need to run with `env` where I remember to set all the necessary environment variables.

I included this bit when I was trying to figure out why things were crashing, it kept varying because a Java version problem and other MAC-related issues.

src/corppa/poetry_detection/passim/create_passim_corpus.py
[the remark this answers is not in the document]
laurejt: I've made a note so we can revisit this in the future.

src/corppa/poetry_detection/passim/get_passim_excerpts.py
laurejt: To better answer the question: there's no real reason why these are two separate scripts. They could probably be combined, but it's worth thinking what output we specifically want/need.

Issue #84, opened 2024-09-11
Integrate passim-related code into `corppa`


mnaydan:
- [ ] Determine which scripts should be incorporated into corppa (and how to integrate organizationally)
- [ ] Incorporate scripts into corppa and write unit tests for adapted code.

Relevant code:
- [ ] Script for processing adjudication data
- [x] #125

rlskoeser:
These tasks were tracked & completed under other issues

PR #130, opened 2024-12-16
Feature/process adjudication data

laurejt:
Ref: #84 

Script for processing prodigy adjudication data files (.jsonl). Also adds project support for type checking via mypy

### DIFF
--- a/src/corppa/poetry_detection/annotation/process_adjudication_data.py
+++ b/src/corppa/poetry_detection/annotation/process_adjudication_data.py
@@ -22,10 +22,10 @@ def get_excerpts(page_annotation: dict[str, Any]) -> list[dict[str, int | str]]:
           poetry detection task.
     """
     excerpts = []
-    page_text = page_annotation["text"]
-    # If spans field is missing, return empty list (i.e., no spans)
+    # Blank pages may not have a text field, so in these cases set to empty string
+    page_text = page_annotation.get("text", "")
     if "spans" not in page_annotation:
-        return excerpts
+        raise ValueError("Page annotation missing 'spans' field")
     for span in page_annotation["spans"]:
         excerpt = {
             "start": span["start"],
--- a/src/corppa/poetry_detection/annotation/process_adjudication_data.py
+++ b/src/corppa/poetry_detection/annotation/process_adjudication_data.py
@@ -1,0 +1,176 @@
+import argparse
+import csv
+import pathlib
+import sys
+from collections.abc import Generator
+from typing import Any
+
+import orjsonl
+from tqdm import tqdm
+from xopen import xopen
+
+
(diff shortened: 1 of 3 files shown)


src/corppa/poetry_detection/annotation/process_adjudication_data.py
```diff
@@ -0,0 +1,176 @@
+import argparse
```
rlskoeser, line 1: Please add a short docstring documenting briefly what this script does, what it is for, and one example of how to run it.

Here's how I ran it locally on the round 2 adjudication data:
```sh
python src/corppa/poetry_detection/annotation/process_adjudication_data.py allround2.jsonl round2_pages round2_excerpts
```

It wasn't obvious to me from the help information that I should have probably named the second param `round2_pages.jsonl` and the second one `round2_excerpts.csv`. I wasn't sure why we need both outputs or what the goals are for the different outputs.

src/corppa/poetry_detection/annotation/process_adjudication_data.py
```diff
@@ -0,0 +1,176 @@
...
+    for excerpt in page_data["excerpts"]:
+        entry = {
+            "page_id": page_data["page_id"],
+            "work_id": page_data["work_id"],
+            "work_title": page_data["work_title"],
+            "work_author": page_data["work_author"],
+            "work_year": page_data["work_year"],
+            "start": excerpt["start"],
+            "end": excerpt["end"],
+            "text": excerpt["text"],
```
rlskoeser, line 76: when/if we come back to this (if we use this script again), let's tweak these fields to match the ones we've agreed on for the poem dataset

rlskoeser (approved):
Looks good, was able to run it on the round2 adjudication data and review the results.  Just needs a bit of documentation on what this is and how to run it, then it can be merged.

src/corppa/poetry_detection/annotation/process_adjudication_data.py
laurejt: Just want to verify that the parser help descriptions are clear. For example, `output_pages` was described as "Filename where extracted pages data (JSONL file) should be written".

rlskoeser: "extracted pages data" was ambiguous to me in this context, I didn't know what it included or how it differed from the other output until poking around in the files a bit.

Maybe something like annotations aggregated/grouped by page  ?

laurejt: Does it make more sense in context now?

laurejt: I'm fine with changing the name later, but not something worth burning more of my R&D time on.

rlskoeser: thanks for updating the help text, it is clearer now

PR #132, opened 2025-01-09
Feature/eval spans

laurejt:
Here is the current code for evaluating span annotations.

The script requires two key files to run, page-level JSONL files that contain reference and system span annotations.

Usage: `python evaluate_poetry_spans ref_jsonl sys_jsonl out.csv`

Note that I'm considering adding reporting for the following:
- number of exact matches
- number of partial matches
- number of missed reference spans
- number of spurious system spans

### DIFF
--- a/src/corppa/poetry_detection/evaluation/evaluate_poetry_spans.py
+++ b/src/corppa/poetry_detection/evaluation/evaluate_poetry_spans.py
@@ -2,6 +2,12 @@
 Evaluate the poetry spans detected by a system against a provided reference set.
 """
 
+import csv
+from collections.abc import Generator
+from pathlib import Path
+
+import orjsonl
+
 
 class Span:
     """
@@ -126,6 +132,7 @@ class PageSystemSpans:
         """
         spans = []
         if page_json["n_spans"] > 0:
+            # TODO: Revisit the format for the system results jsonl
             for span in page_json["poem_spans"]:
                 spans.append(Span(span["page_start"], span["page_end"], span["ref_id"]))
         # Sort spans by primarily by start index and secondarily by end index
@@ -160,6 +167,7 @@ class PageEvaluation:
     Page-level span evaluation.
     """
 
+    page_id: str
     ignore_label: bool  # Flag for ignoring span labels in evaluation
     ref_spans: list[Span]  # Reference spans
     sys_spans: list[Span]  # System spans
@@ -178,6 +186,7 @@ class PageEvaluation:
                 "Reference and system spans must correspond to the same page"
             )
         # Save working input
+        self.page_id = page_ref_spans.page_id
         self.ignore_label = ignore_label
         self.ref_spans = page_ref_spans.spans
         if self.ignore_label:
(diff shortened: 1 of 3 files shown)


laurejt:
> ## [Codecov](https://app.codecov.io/gh/Princeton-CDH/ppa-nlp/pull/132?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Princeton-CDH) Report
> 
> Attention: Patch coverage is `99.45652%` with `2 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 99.64%. Comparing base [(`cb03ab8`)](https://app.codecov.io/gh/Princeton-CDH/ppa-nlp/commit/cb03ab83de981203dd2f5ce658e79ea4a23fc963?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Princeton-CDH) to head [(`718aa4b`)](https://app.codecov.io/gh/Princeton-CDH/ppa-nlp/commit/718aa4b0ef386cc5db31030ce2717642bfdcfc8a?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Princeton-CDH).
> > Report is 2 commits behind head on develop.
> 
> Additional details and impacted files

Looks like Codecov is only checking the test file and not the actual code file.

src/corppa/poetry_detection/evaluation/evaluate_poetry_spans.py
```diff
@@ -0,0 +1,611 @@
...
+        ignore_label: bool,
+    ) -> tuple[list[int | None], list[list[int]]]:
+        """
+        Determines the mappings between the reference and system spans. Each
+        reference span is mapped to at most one system span. Namely, the system
+        span with the highest degree of overlap. Conversely, each system span
+        is mapped to any number of overlapping (but disjoint) reference spans.
+        """
+        ref_to_sys: list[int | None] = [None for _ in ref_spans]
+        sys_to_refs: list[list[int]] = [[] for _ in sys_spans]
```
rlskoeser, line 223: These are lists of numeric ids (or really indices) to match up the spans, right? Why do we need to track them in both directions? Is there a different data structure that would make this easier to work with and understand?

rlskoeser (commented):
I've read through the code and made some comments. I think I'm generally following what you've done, and I'm looking forward to discussing together tomorrow.  I remember you showed me a preliminary diagram of how you were thinking about the span matching - which I can see is happening in the span pairing method; it would be great to diagram that out, maybe on a whiteboard, and then figure out if we can add it to the code documentation or at least a readme.  

I love the span object, and I think the span class at least should be moved out of this script into some place where it will be easier to reuse. Maybe `corppa.poetry_detection.spans` ? IDK if there are other objects we'd put there or what the more general category would be - maybe it eventually it belongs in the same file as the new poem excerpt object.  I do think conceptually the span object behaves like a python `set` in some ways and it could be useful to follow the set interface in some cases (union, intersection, isdisjoint, etc).

At a higher level, I find the terminology of "system" vs "reference" confusing. Perhaps additionally confusing since we've been using reference to talk about our reference corpora and I think this is a different kind of reference. Is this language you're borrowing from somewhere else? If not, maybe we can come up with terms that are a bit easier to differentiate.  At the very least, when we polish this code up together we should have documentation at the top of the file that defines the two terms very clearly.

I tried to run the code locally with the eval data you pointed me to, but I'm getting an error - I tried a multiple different combinations of inputs from the files in that folder, but my confusion about system vs reference may be contributing as well.

I have a lot of suggestions that basically amount to refactoring - when we meet together, let's figure out and agree on what changes we think are needed before we  merge this version of the code into develop, and also what (if any) refactoring we want to do as a second pass after the initial merge.

src/corppa/poetry_detection/evaluation/evaluate_poetry_spans.py
```diff
@@ -0,0 +1,611 @@
...
+            # Edge case to avoid divide by zero error
+            return 1 if not self.ref_spans else 0
+
+        n_retrieved = self._retrieved_count(self.sys_to_refs)
+        relevance_score = self._relevance_score(
+            self.span_pairs, self.ignore_label, partial_match_weight
+        )
+        return relevance_score / n_retrieved
+
+    def recall(self, partial_match_weight: float = 1) -> float:
```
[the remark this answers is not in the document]
laurejt: The main case I see for this is if we wanted to see how the scores with different settings compared. Say, how the results differ when (a) there's no partial match penalty vs (b) some penalty (the case where partial matches are ignored entirely seems meaningless in our particular setting)

src/corppa/poetry_detection/evaluation/evaluate_poetry_spans.py
laurejt: If you've got a favorite graph data structure let me know, this was just the quickest way I could get what I needed.

The reason we need both, is otherwise we need additional logic to determine the unlinked references / system spans.

rlskoeser (approved):
Thanks for writing up all this documentation, it's very helpful.  It's great to have the equations included with the rest. 

I guess I'm a very visual person - I was trying to figure out a mermaid chart that would work to help show the system span with multiple reference spans, to go alongside your table of numbers. I'm glad I did, because I think I understand it better now.  I can open a pull request with that change for you to review after we merge this (and then when you review you can confirm that I do indeed understand it! 😄  )

Issue #129, opened 2024-12-16
write a script to match poem annotation data to reference poem corpus


mnaydan:
- [x] compile full-text into parquet file for quick lookup
- [x] compile metadata in order to look up across sources
- [x] add notes field for initial review
- [x] copy reference data and parquet files to tigerdata 
- [x] take in adjudication data and output poem dataset (ish / subset of fields)
- [x] include method to search by text and get poem metadata
- [x] simple logic to handle more than one match 
- [x] check against manually identified poems from round 1 (preliminary excerpts google sheet, tab is eval_wip - where metadata is filled in)
- [x] run on round 2 adjudication data; check against poems found during adjudication spreadsheet (last set with keywords)

poem dataset fields:
- work_author
- work_title
- work_id
- page_id
- poem_id
- poem_span = the part of the poem that's included on the page
- poem_author
- poem_title
- detection_method = manual annotation
- identification_method = poem lookup by reference conference
- confidence score (nice to have)

If two exact matches, privilege Chadwyck Healey over Poetry Foundation; if two conflicting... TBD. We only care about poem author/title, we don't care about specific editions. (Could think about generating a list of known equivalent ids at some point so that the ids can be more useful; e.g. mint project specific poem ids that aggregrate duplicate editions or copies from different sources)

rlskoeser:
cleaned up and merged

PR #135, opened 2025-01-13
Script to identify poems in annotated excerpts

rlskoeser:
This is the code for #129 

We've already determined that we're not going to do any further refinement or cleanup on the script, since it was an interim step and we either have enough data we need from it, or can run the script in its current form against smaller sets of round 2 annotation data.

Review should be minimal and just flag anything that needs to be documented more clearly for our understanding / memory in future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ __pycache__/
 *$py.class
 
 # project specific
-# - don't check reference poetry data into version control
+# - don't check reference poetry data for identify_poems into version control
 poetry-reference-data/
 
 # C extensions
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# project specific
+# - don't check reference poetry data into version control
+poetry-reference-data/
+
 # C extensions
 *.so
 
--- a/src/corppa/poetry_detection/identify_poems.py
+++ b/src/corppa/poetry_detection/identify_poems.py
@@ -14,6 +14,7 @@ to recompile, rename or remove the parquet files.
 """
 
 import argparse
+import codecs
 import csv
 import logging
 import pathlib
@@ -214,6 +215,7 @@ def _text_for_search(expr):
         # normalize whitespace
         .str.replace_all("[\t\n\v\f\r ]+", " ")  # could also use [[:space:]]
         # replace curly quotes with straight (both single and double)
+        # NOTE: if we revisit, use ftfy for cleanup instead
         .str.replace_all("[”“]", '"')
         .str.replace_all("[‘’]", "'")
         .str.replace_all("ſ", "s")  # handle long s (also handled by unidecode)
@@ -489,8 +491,11 @@ def main(input_file):
     ]
 
     with output_file.open("w", encoding="utf-8") as outfile:
-        # TODO: add byte-order-mark to indicate unicode ?
-        csvwriter = csv.DictWriter(outfile, fieldnames=fieldnames)
+        # add byte-order-mark to indicate unicode
+        outfile.write(codecs.BOM_UTF8)
+        csvwriter = csv.DictWriter(
+            outfile, fieldnames=fieldnames, extrasaction="ignore"
+        )
         csvwriter.writeheader()
 
         for row in input_df.iter_rows(named=True):
@@ -510,7 +515,7 @@ def main(input_file):
                 row["match_count"] = 0
             # write row out with input and any match information found;
             # filter out any fields not in the list of csv field names
-            csvwriter.writerow({k: v for k, v in row.items() if k in fieldnames})
+            csvwriter.writerow(row)
 
     print(f"Poems with match information saved to {output_file}")
     print(
--- a/src/corppa/poetry_detection/identify_poems.py
+++ b/src/corppa/poetry_detection/identify_poems.py
@@ -18,7 +18,12 @@ import csv
 import logging
 import pathlib
 from glob import iglob
-from itertools import batched
+
+try:
+    from itertools import batched
+except ImportError:
+    from more_itertools import batched
+
 from time import perf_counter
 
 import polars as pl
--- a/src/corppa/poetry_detection/identify_poems.py
+++ b/src/corppa/poetry_detection/identify_poems.py
@@ -238,7 +238,7 @@ def multiple_matches(df, search_field):
     match_count = int(df.height)
 
     #  check if both author and title match (ignoring punctuation and case)
-    # TODO: use rapidfuzz here to check author & title are sufficiently similar
+    # TODO: could use rapidfuzz here to check author & title are sufficiently similar
     # e.g. these should be treated as matches but are not currently:
     #    Walter Scott      ┆ Coronach
     #    Walter, Sir Scott ┆ CCLXXVIII CORONACH
@@ -359,9 +359,12 @@ def find_reference_poem(ref_df, input_row, meta_df):
                 match_poem["match_count"] = num_matches
                 return match_poem
 
+    # NOTE: could make configurable to skip fuzzy searches,
+    # or maybe truncate large excepts before running fuzzy search
+
     # if no matches were found yet, try a fuzzy search on the full text
     search_text = unidecode(input_row["search_text"])
-    # TODO: may want to require some minimum length or uniqueness on the search text
+    # NOTE: might want some minimum length or uniqueness on the search text
     logger.info(f"Trying fuzzy match on: {search_text}")
     start_time = perf_counter()
     result = ref_df.with_columns(
@@ -516,7 +519,7 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "input",
-        # TODO: determine minimum required fields; maybe just text?
+        # TODO: determine & document minimum required fields; maybe just text?
         help="csv or tsv file with poem excerpts",
         type=pathlib.Path,
     )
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -3,6 +3,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# project specific
+# - don't check reference poetry data into version control
+poetry-reference-data/
+
```
laurejt, line 9: This seems like a personal setup, rather than something specific to the repo itself

src/corppa/poetry_detection/identify_poems.py
```diff
@@ -0,0 +1,533 @@
+#!/usr/bin/env python
+"""
+
+Script to identify poem excerpts by matching against a local
+collection of reference poems.
+
+Setup:
+
+Download and extract poetry-ref-data.tar.bz2 from /tigerdata/cdh/prosody/poetry-detection
```
laurejt, line 9: What does this correspond to? A particular version of the internet poems, or something else?

src/corppa/utils/filter.py
```diff
@@ -107,8 +107,13 @@ def filter_pages(
...
             # if include filter does not apply, skip this page
-            if not any(page[key] == val for key, val in include_filter.items()):
-                continue
+            try:
+                if not any(page[key] == val for key, val in include_filter.items()):
+                    continue
+            except KeyError as err:
+                raise Exception(
+                    f"Key not found, choose from {','.join(page.keys())}: {err}"
+                )
```
laurejt, line 116: This seems to strict, since it's not a guarantee that all PPA pages have the same fields

src/corppa/utils/filter.py
```diff
@@ -349,7 +354,3 @@ def main():
         # otherwise just report
         else:
             print(f"No pages were selected, output file {output_filepath} is empty")
-
-
-if __name__ == "__main__":
-    main()
```
laurejt, line 355: Why did you remove this?

laurejt: I use the filter script via command line, so this is removing functionality that is in use

src/corppa/utils/filter.py
laurejt, line 1: More explanation is needed for these changes, since they appear to change the existing functionality of the filter script

src/corppa/poetry_detection/identify_poems.py
```diff
@@ -0,0 +1,533 @@
...
+        f"{match_found} excerpts with matches ({match_found / input_df.height * 100:.2f}% of {input_df.height} rows processed)"
+    )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Attempt to identify poem excerpts by matching against reference set"
+    )
+    parser.add_argument(
+        "input",
```
laurejt, line 526: this is a directory, right?

src/corppa/poetry_detection/identify_poems.py
```diff
@@ -0,0 +1,533 @@
...
+        # TODO: determine & document minimum required fields; maybe just text?
+        help="csv or tsv file with poem excerpts",
+        type=pathlib.Path,
+    )
```
laurejt, line 530: I realize we're not cleaning this up, but what about an output directory?

src/corppa/poetry_detection/identify_poems.py
```diff
@@ -0,0 +1,533 @@
...
+                match_poem["match_notes"] = match_poem.pop("notes")
+                row.update(match_poem)
+
+                # update the tally of rows we found matches for
+                match_found += 1
+            else:
+                row["match_count"] = 0
+            # write row out with input and any match information found;
+            # filter out any fields not in the list of csv field names
+            csvwriter.writerow({k: v for k, v in row.items() if k in fieldnames})
```
laurejt, line 513: The filtering of fieldnames shouldn't be necessary, just make DictWriter ignore extra fields by setting `extrasaction='ignore'` in instantiation

src/corppa/poetry_detection/identify_poems.py
```diff
@@ -0,0 +1,533 @@
...
+    fieldnames = input_columns + [
+        "match_count",
+        "match_poem_id",
+        "match_author",
+        "match_title",
+        "match_notes",
+    ]
+
+    with output_file.open("w", encoding="utf-8") as outfile:
+        # TODO: add byte-order-mark to indicate unicode ?
```
laurejt, line 492: I'm pretty sure this is done by setting the encoding to `outfile` to utf-8 (although that is presumably the default)

src/corppa/poetry_detection/identify_poems.py
```diff
@@ -0,0 +1,533 @@
...
+        if non_poetryfoundtn.height == 1:
+            match_poem = non_poetryfoundtn.to_dicts()[0]
+            match_poem["notes"] = (
+                "multiple matches, duplicate author but not title; excluding Poetry Foundation"
+            )
+            return match_poem
+
+
+def fuzzy_partial_ratio(series, search_text):
+    """Calculate rapidfuzz partial_ratio score across for a single input
```
laurejt, line 295:
```suggestion
    """Calculate rapidfuzz partial_ratio score for a single input
```

src/corppa/poetry_detection/identify_poems.py
```diff
@@ -0,0 +1,533 @@
...
+    """Takes a polars expression (e.g. column or literal value) and applies
+    text pattern replacements to clean up to make it easier to find matches."""
+    return (
+        # remove specific punctuation marks in the middle of words
+        expr.str.replace_all(r"(\w)[-'](\w)", "$1$2")
+        # metrical notation that splits words (e.g. sudden | -ly or visit | -or)
+        .str.replace_all(r"(\w) \| -(\w)", "$1$2")
+        # replace other puncutation with spaces
+        .str.replace_all("[[:punct:]]", " ")
+        # remove indent entity in CH (probably more of these...)
```
laurejt, line 212: there are a lot of these, see the chadwyck-healey branch for some of the ones I identified/decoded

src/corppa/poetry_detection/identify_poems.py
```diff
@@ -0,0 +1,533 @@
...
+        .str.replace_all("&indent;", " ")
+        # normalize whitespace
+        .str.replace_all("[\t\n\v\f\r ]+", " ")  # could also use [[:space:]]
+        # replace curly quotes with straight (both single and double)
+        .str.replace_all("[”“]", '"')
+        .str.replace_all("[‘’]", "'")
```
laurejt, line 218: instead of doing all this, it might be better to use ftfy

src/corppa/utils/filter.py
laurejt: I realize it's not needed to run `corppa-filter-corpus`, but it's not clear we need to remove these lines either.

src/corppa/utils/filter.py
laurejt: Although I guess this error would have occurred anyway

laurejt (commented):
Given there's no intent to refine or clean up this script, the core script seems fine. However, given the found issue with the utility filter script, it seems instead of throwing a keyerror, that the script should ignore pages with these missing keys, since PPA pages are not guaranteed to have the same fields.

I believe I may have hit the wrong button for my review, I believe this should *not* be merged until the fix to `filter.py` is made.

src/corppa/utils/filter.py
rlskoeser: I don't remember doing it, I think it must have been unintentional - thanks for catching, I'll restore

src/corppa/poetry_detection/identify_poems.py
rlskoeser: I aggregated all of our existing sources in a preferred order (internet poems, CH, poetry foundation)

src/corppa/poetry_detection/identify_poems.py
```diff
@@ -0,0 +1,533 @@
...
+REF_DATA_DIR = pathlib.Path("poetry-reference-data")
+TEXT_PARQUET_FILE = REF_DATA_DIR / "poems.parquet"
+META_PARQUET_FILE = REF_DATA_DIR / "poem_metadata.parquet"
+# csv files to supplement .txt files
+POETRY_FOUNDATION_CSV = REF_DATA_DIR / "poetryfoundationdataset.csv"
+CHADWYCK_HEALEY_CSV = REF_DATA_DIR / "chadwyck_healey_metadata.csv"
+# define source ids to ensure we are consistent
+SOURCE_ID = {
+    "Poetry Foundation": "poetry-foundation",
+    "Chadwyck-Healey": "chadwyck-healey",
```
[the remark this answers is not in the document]
rlskoeser: I don't remember why I didn't do one for the internet poems - maybe because it has no spreadsheet and I wasn't reconciling two different things? Or I was just inconsistent

src/corppa/poetry_detection/identify_poems.py
rlskoeser: I tested it, it doesn't. But it's easy to add, I'm going to add the line now.

src/corppa/poetry_detection/identify_poems.py
rlskoeser: I wasn't familiar with that option, thank you

src/corppa/utils/filter.py
rlskoeser: Yes, I think I ran into a problem and this let me at least see what the error was in a more useful way.  I don't remember now why this change was in this pr.  I'm going to leave it for now and we can revisit if/when we run into it so we can decide what behavior we want.

src/corppa/poetry_detection/identify_poems.py
rlskoeser: ah, yeah, I should have. I made a note that if we revisit we should switch to that

Issue #139, opened 2025-01-28
Add Python script for running passim


laurejt:
Create a Python script for running passim. The goal with this script is to (1) allow for input file validation and (2) more easily pass along optional arguments to passim for optimizing passim.

PR #140, opened 2025-01-28
Added python script to run passim & its unit tests

laurejt:
ref: #139


rlskoeser (approved):
This seems fine. Should the older `run_passim.sh` be removed?

I was able to run the python script locally; I just ran it on some data I had in my working directory. I didn't get any output to the terminal but I guess that's because I didn't request verbose output.

Fine with me to merge, especially since you're already using it.

Issue #142, opened 2025-02-05
merge Chadwyck-Healey text generated from tml file code into Chadwyck-Healey feature branch as a pull request


mnaydan:
Wouter steps:
- [x] Wouter: commit it to branch directly
- [x] Wouter: make a pull request against develop
- [x] tag Laure for review in GitHub

Laure steps:
- [x] Review pull request
- [x] Determine if any of the 90 skipped files are an issue (no)
- [x] Revise code as needed
- [x] Optionally add unit tests
- [x] tag Rebecca to review the updated GitHub PR

revisions to merge PR:
- [x] make text encoding check optional, off by default and assume Latin 1 (leave the logic so we can use/improve if needed)
- [x] add a metadata only flag to allow generating csv without parsing text
- [x] minimal fixes (or remove) `get_tag_stats` script so it runs

PR #157, opened 2025-02-10
Feature/chadwyck healey

WHaverals:
#142 

I built a new TML-file parser! For multiple reasons:

1. Most importantly, these files are weird to say the least -- not well-formed XML, nor SGML, nor HTML...
2. the `plaintext` version of the corpus (we received from Mark Algee-Hewitt) contained lot of weirdness... e.g.:

- many files still contain waht looks like HTML tags
- other encoding issues
- lots of copyright notices
- extremely weird indentation (which causes some files to be well over 10MB, one is even 335MB -- simply because each newline is a tab, see e.g. `Z200474224`)
- too many blank lines too...

3. Mark also sent us a metadata file (`POETRY_CORPUS.xlsx`), which is a mess too, e.g.:

- all the translators are forgotten; and these are sometimes just really important! (this is a big issue for me - e.g. Guy Fawkes translated Theocritus, and this is not mentioned anywhere in the xlsx file...)
- encoding errors (e.g. `A Sash for Wu YÃ¼n, â€œFifteen stars fence polar fire ...â€`, etc.)

So in short, both the the plaintext files and the metadata file need to be cleaned up.
I went back to the original TML files, and tried to parse them - to the best of my ability, to create a clean txt corpus.

The relevant files in this pull request are:
- `parse_tml.md` (~ description of the main logic that was followed and implemented; there are many edge cases)
- `entity_map.py` (~ I made an attempt to map certain exotic entities (XML codes, HTML codes, etc.) that need to be resolved; the most important (frequent ones) should be resolved -- the ones remaining are part of the really long tail of singletons)
- `tml_parser.py` (~ the main script!)

cmd to run the script: `python tml_parser.py --input_dir "tml" --output_dir "tml_parsed"`

optional args: 
`--num_files 1000`  (leave out to process all files)
`--verbose` (for verbose output)


src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -0,0 +1,689 @@
...
+        Returns:
+            str: Text with entities replaced.
+        """
+        if not text:
+            return ""
+
+        # replace all known entities
+        for entity, replacement in self.entity_map.items():
+            # handle both with and without semicolon
+            entity_with_semicolon = entity if entity.endswith(";") else f"{entity};"
```
laurejt, line 106: Does this actually happen? Will this replace any text that should not be modified?

src/corppa/poetry_detection/chadwyck_healey/entity_map.py
```diff
@@ -0,0 +1,97 @@
...
+    "&wblank;": "\u2003",  # EM SPACE
+    "&point;": ".",
+    "&supere;": "ᵉ",  # U+1D49
+    "&ebar;": "ē",  # U+0113
+    "&obar;": "ō",  # U+014D
+    "&supert;": "ᵗ",  # U+1D57
+    # i'm not sure about the next one (https://www.compart.com/en/unicode/U+2423)
+    # it might be a blank line instead of a space
+    # tread cautiously, though, because it appears in metadata fields
+    # if we'd just replace it with a blank line, we'd create a new row in the metadata...
```
laurejt, line 16: This won't actually break the metadata file, if the character is properly escaped (which the `csv` package will do)

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -0,0 +1,689 @@
...
+            - Strip leading/trailing whitespace.
+            - Replace known entities.
+        Parameters:
+            text (str): The raw text to be cleaned.
+        Returns:
+            str: Cleaned and normalized text.
+        """
+        if not text:
+            return ""
+        text = re.sub(r"\s+", " ", text)
```
laurejt, line 127: This will remove all newlines... which seems too aggressive

src/corppa/poetry_detection/chadwyck_healey/build_poem_corpus.py
laurejt, line 1: This is just a copy of `tml_parser.py`, so I'll delete it in a subsequent commit

src/corppa/poetry_detection/chadwyck_healey/entity_map.py
laurejt, line 1: This is really part of the TML Parser, so I'll add move the mapping there as a global variable

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -0,0 +1,689 @@
...
+        """
+        if not text:
+            return ""
+
+        # replace all known entities
+        for entity, replacement in self.entity_map.items():
+            # handle both with and without semicolon
+            entity_with_semicolon = entity if entity.endswith(";") else f"{entity};"
+            text = text.replace(entity_with_semicolon, replacement)
+            text = text.replace(entity.rstrip(";"), replacement)
```
laurejt: Why are ending semicolons being removed from the text?

laurejt: One issue that can arise here, it is very important what order the entities are processed since there's nothing that prevents entity names from being prefixes of another.

src/corppa/poetry_detection/chadwyck_healey/build_poem_corpus.py
```diff
@@ -1,405 +1,689 @@
...
         """
-        metadata = {}
-        tag_types = Counter(
-            [tag.name for tag in self.head.find_all(True, recursive=False)]
-        )
-        for tag in self.head.find_all(True, recursive=False):
-            if tag.name == "body":
-                # This shouldn't happen, unless there's a formatting issue
-                self.print_log(f"Encountered body tag in head tag, skipping")
+        encodings = ["utf-8", "latin-1", "windows-1252", "cp1252", "ascii"]
```
laurejt: since ascii is a subset of utf-8; the ascii encoding is redundant

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -0,0 +1,689 @@
...
+                )
+
+            return metadata, poetry_text
+
+        except Exception as e:
+            print(f"Error processing {file_path}:")
+            print(f"Error type: {type(e).__name__}")
+            print(f"Error message: {str(e)}")
+            print(traceback.format_exc())
+            return None, None
```
laurejt: This means we don't actually parse all files! What are we missing in these cases?

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -0,0 +1,689 @@
...
+        if failed_files:
+            print(f"Failed to process {len(failed_files)} files:")
+            for fname in failed_files:
+                print(f"  - {fname}")
+
+        # print a list of TML files that contained only a figure
+        if self.figure_only:
+            print("\nThe following files contained only a figure (no text):")
+            for f in self.figure_only:
+                print("  -", f)
```
laurejt, line 523: `self.figure_only` is never updated... is this intentional?

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
laurejt: This seems to be a byproduct of using the `html.parser` to parse these non-HTML files. Switching to `lxml` should fix this issue.

laurejt:
@rlskoeser Note that the parsing is quite ad hoc and is likely to miss text that has a different structure. For example, it looks like speaker div tags are ignored and text that for one reason or another fall outside of tags can be lost as is the case with Z300491786.tml

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -0,0 +1,970 @@
...
+         - Cleaning the final output text.
+        Parameters:
+            element (Tag): A BeautifulSoup element.
+        Returns:
+            str: The cleaned text content.
+        """
+        if not element:
+            return ""
+
+        # convert italic spans (e.g., <span class="italic">word</span> => word)
```
rlskoeser, line 393: It seems like it would be cleaner and easier to use beautifulsoup strings instead of text, but not sure we want to change that now.

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -0,0 +1,970 @@
...
+                elif lname == "Anon.":
+                    anon_author = author_info
+
+            # assign extracted metadata to final dictionary
+            if anon_author:
+                metadata["author_lastname"] = "Anon."
+                metadata["author_firstname"] = anon_author.get("firstname", "")
+                metadata["author_birth"] = anon_author.get("birth", "")
+                metadata["author_death"] = anon_author.get("death", "")
+                metadata["author_period"] = anon_author.get("period", "")
```
rlskoeser, line 859: It doesn't seem like this actually needs to be different logic than orig author

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -0,0 +1,970 @@
...
+            span.replace_with(uppercase_text)
+
+        # now get the text!
+        return self.clean_text(element.get_text())
+
+    def print_warning(self, message: str) -> None:
+        warning = f"Warning: {message}"
+        if self.current_file:
+            warning += f" for {self.current_file}"
+        print(warning)
```
rlskoeser, line 410: glad to see that there's a way to provide filename as context for warnings encountered during parsing

rlskoeser (commented):
Trying to keep my comments at fairly high level based on my understanding of our priorities for this script
- switching from subprocess + file to python-magic would improve efficiency and be a light lift (I can contribute if we decide it's worth investing in it); it does require libmagic to be installed; a better solution would be a one-time option to sanitize (and possible filter) the input files so it doesn't have to be done every time
- it seems like it would be useful to add options for a metadata only mode, to generate the spreadsheet without doing all the text parsing; the way the code is set up that should be pretty easy to integrate
- in future it might be nice to add an option to transform specific ids/filenames, so if we have a problem with a specific text or set of texts we don't have to regenerate everything; we could also add a check to skip text files that already exist in the output directory unless a flag is specified to overwrite/regenerate

I was able to run the script locally; fixed one variable that didn't work when specifying a limit (I think it just got missed on a refactor).  I tried to add files to the `figure_only` list in two different places, but it didn't work - those files were reported as failed to parse.

The parsing seems to be very complicated and seems to some redundancy; I think we can get beautifulsoup to do more of the work for us.   My approach would be to start writing unit tests for the different cases we know we need to handle and expected output and then start revising the code, but I'm not clear on how much of a priority that is right now.

rlskoeser:
> @rlskoeser Note that the parsing is quite ad hoc and is likely to miss text that has a different structure. For example, it looks like speaker div tags are ignored and text that for one reason or another fall outside of tags can be lost as is the case with Z300491786.tml

Coming back to the main page and saw this comment again. @laurejt  how concerned are you about the parsing, and how much do you think we should invest in improving it? Would it make sense to do as a second pass after this initial script is merged in?

laurejt:
> > @rlskoeser Note that the parsing is quite ad hoc and is likely to miss text that has a different structure. For example, it looks like speaker div tags are ignored and text that for one reason or another fall outside of tags can be lost as is the case with Z300491786.tml
> 
> Coming back to the main page and saw this comment again. @laurejt how concerned are you about the parsing, and how much do you think we should invest in improving it? Would it make sense to do as a second pass after this initial script is merged in?

I think we just need to deal with this issue later. I think we need to determine whether we have the bandwidth to triage how bad the parsing is failing, and what if any that we need to correct. It's just not possible in this step, since there's no infrastructure for testing what is/is not being parsed correctly.

src/corppa/poetry_detection/chadwyck_healey/entity_map.py
```diff
@@ -0,0 +1,97 @@
+# entity map based on the most common entities found in the corpus
+
+entity_map = {
+    "&indent;": "\t",
+    "&yogh;": "ʒ",  # U+0292
+    "&ast;": "*",
+    "&wblank;": "\u2003",  # EM SPACE
+    "&point;": ".",
+    "&supere;": "ᵉ",  # U+1D49
```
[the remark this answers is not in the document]
laurejt: Since it's in the code right now I say we don't do anything. My planned way of dealing with this downstream is to use `ftfy` with NFKC normalization which will get rid of these kinds of characters (e.g., `™` --> `TM` and `H₂O` --> `H2O` and so on).

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -0,0 +1,970 @@
...
+    elif " Non-ISO extended-ASCII text" in p.stdout:
+        # Assume this is Windows-1252
+        return "cp1252"
+    elif " UTF-8 text" in p.stdout:
+        return "utf-8"
+    elif " ASCII text" in p.stdout:
+        # Treat ASCII as UTF-8
+        return "utf-8"
+    else:
+        raise ValueError(f"Unknown encoding: {p.stdout}")
```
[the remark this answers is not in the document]
laurejt: I think we should just axe all of this, since all of the files can be opened as `latin-1`

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
laurejt: Yeah, I spent a bit of time adding this logic because otherwise it was impossible to know what file without using the verbose output printing the name of each file in turn

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -759,18 +768,28 @@ def process_directory(self, num_files: Optional[int] = None):
...
-                    # write poetry to text file
-                    output_file = self.output_dir / f"{file_path.stem}.txt"
-                    with open(output_file, "w", encoding="utf-8") as f:
-                        f.write(poetry_text)
-                    n_processed += 1
+                if self.metadata_only:
+                    # Only an attempt at extracting metadata is made
+                    if metadata:
+                        writer.writerow(metadata)
+                        n_processed += 1
```
rlskoeser, line 775: It would be better if the metadata output logic isn't duplicated; I started trying to do that, but it seems that it might  have other implications (what counts as success when you're outputting both?) and might not be any less complex.  

Here's what I drafted trying to figure it out, in case it's useful. (I couldn't make a suggestion because it spans some deleted lines)
```python
processed = False
# always export the metadata if we have it
if metadata:
    # write metadata to CSV
    writer.writerow(metadata)
    processed = True
# if we have text (i.e., not running in metadata only mode
# and text was successfully extracted)
if poetry_text and not self.metadata_only:
    # write poetry to text file
    output_file = self.output_dir / f"{file_path.stem}.txt"
    with open(output_file, "w", encoding="utf-8") as f:
        f.write(poetry_text)
    processed = True

# if we don't have data we expect for this mode, consider it failed
if not metadata or not (poetry_text and not self.metadata_only):
    failed_files.append(file_path.name)
n_attempted += 1
if processed:
    n_processed += 1
```

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
```diff
@@ -682,13 +688,16 @@ def process_file(self, file_path: Path):
...
-            # TODO: This is definitely not the right parser to use!
-            soup = BeautifulSoup(content, "html.parser")
+            soup = BeautifulSoup(content, "lxml")
+
             # extract metadata from the header section
             metadata = self.extract_metadata(soup)
             metadata["filename"] = file_path.name
+
+            if self.metadata_only:
+                return metadata, None
```
rlskoeser, line 698: ok, reasonable for the current structure. suggest adding a comment
```suggestion

            # in metadata-only mode, bail out before text extraction logic, returning None for poetry_text
            if self.metadata_only:
                return metadata, None
```

rlskoeser (approved):
Revisions look reasonable, let's get it merged!

src/corppa/poetry_detection/chadwyck_healey/tml_parser.py
laurejt: I'll add the change. This speaks of a much larger structural issue in that the metadata and text extraction should be separate. There's an unnecessary dependency, but that's not a minimal change.

PR #158, opened 2025-02-10
Add basic poem excerpt data type

laurejt:
Ref: #144

### DIFF
--- a/src/corppa/poetry_detection/poem_excerpts.py
+++ b/src/corppa/poetry_detection/poem_excerpts.py
@@ -32,13 +32,17 @@ class PoemExcerpt:
     def __post_init__(self):
         # Check PPA span indices
         if self.ppa_span_end <= self.ppa_span_start:
-            raise ValueError("PPA span's start index must be less than end index")
+            raise ValueError(
+                f"PPA span's start index {self.ppa_span_start} must be less than its end index {self.ppa_span_end}"
+            )
         # Check that both reference span indicies are set or unset
         if (self.ref_span_start is None) ^ (self.ref_span_end is None):
             raise ValueError("Reference span's start and end index must both be set")
         # Check reference span indices if set
         if self.ref_span_end is not None and self.ref_span_end <= self.ref_span_start:
-            raise ValueError("Reference span's start index must be less than end index")
+            raise ValueError(
+                f"Reference span's start index {self.ref_span_start} must be less than its end index {self.ref_span_end}"
+            )
 
     def to_dict(self) -> dict[str, Any]:
         """
--- a/src/corppa/poetry_detection/poem_excerpts.py
+++ b/src/corppa/poetry_detection/poem_excerpts.py
@@ -3,9 +3,10 @@ Custom data type for poetry excerpts identified with the text of PPA pages.
 """
 
 from dataclasses import dataclass
+from typing import Any, Optional
 
 
-@dataclass
+@dataclass(kw_only=True)
 class PoemExcerpt:
     """
     A poem excerpt data type representing an identified poem excerpt within a
@@ -13,17 +14,38 @@ class PoemExcerpt:
     """
 
     # PPA page related
-    ppa_page_id: str
+    page_id: str
     ppa_span_start: int
     ppa_span_end: int
     ppa_span_text: str
     # Reference poem related
     poem_id: str
     ref_corpus: str
-    ref_span_start: Optional[int]
-    ref_span_end: Optional[int]
-    ref_span_text: Optional[str]
+    ref_span_start: Optional[int] = None
+    ref_span_end: Optional[int] = None
+    ref_span_text: Optional[str] = None
     # Methods & additional notes
     detection_methods: set[str]
     identification_methods: set[str]
-    notes: Optional[str]
+    notes: Optional[str] = None
+
+    def to_json(self) -> dict[str, Any]:
+        """
+        Returns a JSON-style dict of the poem excerpt
+        """
+        json_dict = {
+            "page_id": self.page_id,
+            "poem_id": self.poem_id,
+            "ppa_span_start": self.ppa_span_start,
+            "ppa_span_end": self.ppa_span_end,
+            "ppa_span_text": self.ppa_span_text,
+            "poem_id": self.poem_id,
+            "ref_corpus": self.ref_corpus,
+            "ref_span_start": self.ref_span_start,
+            "ref_span_end": self.ref_span_end,
+            "ref_span_text": self.ref_span_text,
+            "detection_methods": list(self.detection_methods),
+            "identification_methods": list(self.identification_methods),
+            "notes": self.notes,
+        }
+        return json_dict
--- a/src/corppa/poetry_detection/poem_excerpts.py
+++ b/src/corppa/poetry_detection/poem_excerpts.py
@@ -1,0 +1,29 @@
+"""
+Custom data type for poetry excerpts identified with the text of PPA pages.
+"""
+
+from dataclasses import dataclass
+
+
+@dataclass
+class PoemExcerpt:
+    """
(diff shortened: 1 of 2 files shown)


rlskoeser (approved):
Great! We might want to add field-level docstrings at some point, but given that we never merged in our preliminary sphinx docs setup, I think let's merge this and move on to the next steps.

test/test_poetry_detection/test_poem_excerpts.py
```diff
@@ -0,0 +1,51 @@
...
+        result = poem_excerpt.to_json()
+        assert result == expected_result
+
+        # With optional fields
+        poem_excerpt.ref_span_start = 2
+        poem_excerpt.ref_span_end = 3
+        poem_excerpt.ref_span_text = "ref_text"
+        poem_excerpt.notes = "notes"
+
+        expected_result |= {
```
rlskoeser, line 43: cool syntax, I wasn't aware of this either

src/corppa/poetry_detection/poem_excerpts.py
```diff
@@ -0,0 +1,51 @@
...
+    ref_corpus: str
+    ref_span_start: Optional[int] = None
+    ref_span_end: Optional[int] = None
+    ref_span_text: Optional[str] = None
+    # Methods & additional notes
+    detection_methods: set[str]
+    identification_methods: set[str]
+    notes: Optional[str] = None
+
+    def to_json(self) -> dict[str, Any]:
```
rlskoeser, line 32: How is this different from the dataclass `asdict` method?

If we're customizing, can we omit the unset values?

Also, could we rename the method to `to_dict` or `as_dict` ? Calling it `to_json` seems misleading

src/corppa/poetry_detection/poem_excerpts.py
```diff
@@ -0,0 +1,51 @@
...
+    PPA page
+    """
+
+    # PPA page related
+    page_id: str
+    ppa_span_start: int
+    ppa_span_end: int
+    ppa_span_text: str
+    # Reference poem related
+    poem_id: str
```
rlskoeser, line 22: Poem id needs to be optional, we won't always be able to identify the poem (we may have marked this incorrectly on the whiteboard)

src/corppa/poetry_detection/poem_excerpts.py
laurejt: I had missed that there was an `asdict` method provided by dataclass. Thanks for mentioning this!

My guess is that the big change is converting the sets to lists since JSON doesn't support lists... but maybe I'm missing some functionality of the `orjsonl` package?

Happy to rename the method to `to_dict`.

laurejt: We can make the unset values optional, but it is annoying to have to check for the existence of these fields on the parsing side; especially for the large number that should generally be set (i.e., all the poem identification fields)

rlskoeser: Right, we do need to convert sets to lists. Maybe use `asdict` and then only customize the result where needed?es

Regarding skipping optional fields: I was just thinking that the json files will be much smaller if we can omit them.

Can we use this class for parsing, at least if/when we're using python?

src/corppa/poetry_detection/poem_excerpts.py
rlskoeser: I think we should include unidentified poems in the lower level dataset, although certain aggregations will inevitably drop them.  But if that makes a lot of extra work we may need to reconsider.

I see what you mean about how that impacts other fields, though. Maybe I'm over complicating this. Can we run with this version and then revise (with validation) when/if we hit a subset of excerpts that requires it?

laurejt: That sounds good to me. All of these things are worth talking about more, but you're right in pointing out that we don't need to deal with it quite yet.

src/corppa/poetry_detection/poem_excerpts.py
laurejt: Sure, we can omit them for now; if we compress the files, I assume there wouldn't be much difference.

I assume we can probably use this class for parsing... but I haven't gotten to that part of development yet, so I'm not sure what functionality we'd exactly be incorporating.

rlskoeser:
@laurejt do we get the validation for free when we incorporate your span class? If so, I think you should hold off on implementing any validation now

src/corppa/poetry_detection/poem_excerpts.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        json_dict = asdict(self)
+        json_dict["detection_methods"] = list(self.detection_methods)
+        json_dict["identification_methods"] = list(self.identification_methods)
+
+        # Remove optional fields if set to None
+        for field in ["ref_span_start", "ref_span_end", "ref_span_text", "notes"]:
+            if json_dict[field] is None:
+                del json_dict[field]
+        return json_dict
```
rlskoeser, line 45: fwiw, I would write it this way, which I think is more pythonic and is agnostic about which fields are optional:

But your version is fine for now!

```suggestion
        return {k:v for k, v in json_dict.items() if v is not None}
```

rlskoeser (approved):
Your revisions look great! I vote we get this merged and hold off on the other revisions and validation we've discussed in the comments.

src/corppa/poetry_detection/poem_excerpts.py
laurejt: My one worry here is we ever have a required field that can be None... this would remove those ones. Which is fine as long as we explicitly document this behavior. It was quite a surprise when I learned that the `text` field is not required for the ppa page jsonl structure

laurejt:
> @laurejt do we get the validation for free when we incorporate your span class? If so, I think you should hold off on implementing any validation now

Mostly? Although we'd still need to handle the case where only one of the reference span indices is specified. I already committed the change, so it matters less. It's easy enough to remove the related test cases if/when the Span class is ported

src/corppa/poetry_detection/poem_excerpts.py
```diff
@@ -0,0 +1,56 @@
...
+    def __post_init__(self):
+        # Check PPA span indices
+        if self.ppa_span_end <= self.ppa_span_start:
+            raise ValueError("PPA span's start index must be less than end index")
+        # Check that both reference span indicies are set or unset
+        if (self.ref_span_start is None) ^ (self.ref_span_end is None):
+            raise ValueError("Reference span's start and end index must both be set")
+        # Check reference span indices if set
+        if self.ref_span_end is not None and self.ref_span_end <= self.ref_span_start:
+            raise ValueError("Reference span's start index must be less than end index")
```
rlskoeser, line 41: Could be helpful to include the values in the error message. (I guess this change will require revising the test message)

```suggestion
            raise ValueError(f"Reference span's start index {self.ref_span_start} must be less than end index {self.ref.span_end}")
```

rlskoeser (approved):
Validation looks good.

src/corppa/poetry_detection/poem_excerpts.py
laurejt: I'll make a commit that includes this suggestion and updates the test accordingly. I'll do the same for the page span indices as well.

rlskoeser:
@laurejt the conversations on this PR have been so useful in terms of thinking through our data structure, what kind of documentation is important, validation, what kind of variation we want to allow! Should we make an issue to gather the possible refinements we discussed but didn't do yet? Or maybe it's fine to leave them as comments on this PR since we'll be working on it this week and next and will probably remember in that time frame...

laurejt:
@rlskoeser Perhaps we can gather the refinements into a draft issue for now? It's possible they might get rolled into some of the ongoing development; but I think it'd be valuable to consolidate all the different points raised in this PR

PR #159, opened 2025-02-11
Configure test paths for pytest to speed up test collection

rlskoeser:
I noticed when trying to run tests on the poem excerpt branch that collecting tests was really slow. Specifying the test paths as a pytest config option speeds it up significantly.


laurejt (approved):
Good catch! I had just encountered the slow down myself.

PR #163, opened 2025-02-17
Add chart to show how system span is split for multiple reference spans

rlskoeser:
Adds a visual representation of how spans are split into subspans pairs for evaluation using a mermaid chart.

I couldn't figure out how to add comments to explain what's going on, but I'm using column width and spaces to try to convey how these spans overlap.

You can preiew the chart here: https://github.com/Princeton-CDH/ppa-nlp/tree/sysrefspan-chart/src/corppa/poetry_detection/evaluation

I think I may have switched system and reference spans - although I was trying to follow the numeric example you provided. Please confirm and let me know if you see anything else that should be revised and I'll address.

### DIFF
--- a/src/corppa/poetry_detection/evaluation/README.md
+++ b/src/corppa/poetry_detection/evaluation/README.md
@@ -75,7 +75,26 @@ $$
 \left(r_{k_{start}}, s_{end} \right)
 $$
 
+
+```mermaid
+block-beta
+columns 12
+space:1
+Sys["System Span"]:9
+space:2
+Ref1["Ref Span 1"]:3
+space:1
+Ref2["Ref Span 2"]:3
+Ref3["Ref Span 3"]:4
+space:2
+SysSub1["System Sub 1"]:3
+SysSub2["System Sub 2"]:3
+SysSub3["System Sub 3"]:3
+```
+
 **Example.**
+
+
 If we have the following four reference spans $(394, 512), (516, 557), (563, 633), (637, 675)$ mapped
 to the system span $(389, 678)$, this will result in the following working reference-system span pairs:
 
--- a/src/corppa/poetry_detection/evaluation/README.md
+++ b/src/corppa/poetry_detection/evaluation/README.md
@@ -61,9 +61,9 @@ b["Span b"]:7
 space:2
 overlap:4
 
-style a fill:green
-style b fill:blue
-style overlap fill:purple
+style a fill:#147416
+style b fill:#11149C
+style overlap fill:#481E66
 ```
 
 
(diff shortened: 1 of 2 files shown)


rlskoeser:
> Overall, this looks good to me. One suggest change is that the final reference span is ends before the system span to show that the last system subspan will end at the system's end not the paired references.

Ah, that's helpful, thank you. I've re-ordered so the full system and split system spans are in adjacent rows and I added a color to connect them. How does the revised version look to you?

rlskoeser:
@laurejt should I renumber spans in the chart to start with 0 for consistency with your equation?

rlskoeser:
@laurejt I added another diagram to illustrate the span overlap, so tagged you to re-review. I also fixed a test that was flagged by the test coverage check.

src/corppa/poetry_detection/evaluation/README.md
```diff
@@ -75,7 +94,33 @@ $$
...
+SysSub2["System Sub 2"]:3
+SysSub3["System Sub 3"]:4
+
+style Sys fill:#484
+style SysSub1 fill:#484
+style SysSub2 fill:#484
+style SysSub3 fill:#484
+
+```
+
```
laurejt, line 120: I find the added spacing between the system sub-spans confusing, since they should be contiguous. It also deceptively makes it look like the second system sub-span ends at the end index of the second reference span rather than the starting index of the third reference span.

laurejt (changes requested):
I like the addition of both figures, but I think there are a few major changes we need to make.

- Change up the fill colors so that the text is legible
- Change to 0-index spans on the second figure
- Remove/update spacing between spans to mitigate confusion about the sub-span indices

src/corppa/poetry_detection/evaluation/README.md
rlskoeser: I'm confused by your comment - it makes me wonder if you're seeing something different than I am.  There aren't any spaces between the system subspans. I adjusted after your first comment to move the last reference span to make it clearer that the last subspan ends at the end of the system span, not the reference span.

laurejt: There is some horizontal spacing between each span, which I'm assuming was to show intentional gaps between each reference span. However, there should be no gapping between the system-spans; since they are contiguous slices.

Below is a quick sketch trying to highlight what I mean. "System Sub 1" looks like it ends at the  end of "Ref Span 1", rather than the "Ref Span 2 start"

gap_issues

laurejt:
> > For the new diagram is it possible to get nicer colors? It's really difficult to read the the text within those spans because of the fill color.
> 
> Sure, easy to change. Could you suggest colors? Or we could go lighter and change the text color.

Yes, lighter seems find. It's just that the colors are really dark, so the text is impossible to read for the green/purple/blue ones.

laurejt:
@rlskoeser The colors as of the last update still seem too dark. I can't read the text. Could we just use something like LightGreen (#90EE90), Lavender (#E6E6FA), and LightCyan (#E0FFFF); anything that's just a much lighter color.

rlskoeser:
> @rlskoeser The colors as of the last update still seem too dark. I can't read the text. Could we just use something like LightGreen (#90EE90), Lavender (#E6E6FA), and LightCyan (#E0FFFF); anything that's just a much lighter color.

I only just figured out, with this comment and your screenshot, that we are dealing with a light mode / dark mode issue! I will need to set the foreground color as well as the background. Do you prefer light mode colors?

laurejt (approved):
Thanks for figuring out the root cause of the color issue!

Issue #151, opened 2025-02-10
update adjudication extraction to strip trailing/leading whitespace and update indices


mnaydan:
- [x] Preserve original extracted data: `tigerdata/cdh/prosody/ppa-prodigy/adjudicated_excerpts/raw_excerpts`
- [x] Updated Python script to update excerpts
- [x] Extract clean excerpt data from adjudication data and upload to tigerdata: `tigerdata/cdh/prosody/ppa-prodigy/adjudicated_excerpts/cleaned_excerpts
- [x] Updated [Adjudicated Excerpts](https://docs.google.com/spreadsheets/d/1SCAdAp1KKUxwIjI0pnLS0iq-RyN2ij1KMSLJmCaNvk8/edit?gid=1796663720#gid=1796663720) spreadsheet

laurejt:
Original version of the files are located on tigerdata at `/tigerdata/cdh/prosody/ppa-prodigy/adjudication/original_processed_data`

laurejt:
New versions of extracted adjudicated data are located on tigerdata at `cdh/prosody/ppa-prodigy/adjudication/excerpt_data`

laurejt:
Note that the spreadsheet was missing the excerpts from round1 that were not in the preliminary set; this has been corrected.

PR #166, opened 2025-02-18
Add logic to strip leading and trailing whitespace from extracted spans

laurejt:
ref: #151

### DIFF
--- a/src/corppa/poetry_detection/annotation/process_adjudication_data.py
+++ b/src/corppa/poetry_detection/annotation/process_adjudication_data.py
@@ -32,6 +32,24 @@ from tqdm import tqdm
 from xopen import xopen
 
 
+def clean_excerpt(span: dict[str, int | str]):
+    """
+    Clean excerpt so that all leading and trailing whitespace is removed.
+    """
+    updated_span = span.copy()
+    # Remove any leading whitespace
+    ldiff = len(span["text"]) - len(span["text"].lstrip())
+    if ldiff:
+        updated_span["start"] = span["start"] + ldiff
+        updated_span["text"] = updated_span["text"][ldiff:]
+    # Remove any trailing whitespace
+    rdiff = len(span["text"]) - len(span["text"].rstrip())
+    if rdiff:
+        updated_span["end"] = span["end"] - rdiff
+        updated_span["text"] = updated_span["text"][:-rdiff]
+    return updated_span
+
+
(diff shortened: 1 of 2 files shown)


src/corppa/poetry_detection/annotation/process_adjudication_data.py
```diff
@@ -32,6 +32,24 @@
 from xopen import xopen
 
 
+def clean_excerpt(span: dict[str, int | str]):
+    """
+    Clean excerpt so that all leading and trailing whitespace is removed.
```
rlskoeser, line 37: Might be nice to document what the span dict fields are (maybe it's obvious in context of this file but I'd forgotten and we have other spans)
```suggestion
    Clean excerpt so that any leading and trailing whitespace is removed, and update start and end values to match the cleaned text. Expects a dictionary with start and end indices and text content.
```

src/corppa/poetry_detection/annotation/process_adjudication_data.py
```diff
@@ -32,6 +32,24 @@
...
+    # Remove any leading whitespace
+    ldiff = len(span["text"]) - len(span["text"].lstrip())
+    if ldiff:
+        updated_span["start"] = span["start"] + ldiff
+        updated_span["text"] = updated_span["text"][ldiff:]
+    # Remove any trailing whitespace
+    rdiff = len(span["text"]) - len(span["text"].rstrip())
+    if rdiff:
+        updated_span["end"] = span["end"] - rdiff
+        updated_span["text"] = updated_span["text"][:-rdiff]
```
rlskoeser, line 49: this is clever

rlskoeser (approved):
Logic is good, unit test includes good set of test cases.

Any documentation needed or next steps implied? How are we going to make use of this code change?

src/corppa/poetry_detection/annotation/process_adjudication_data.py
laurejt: Sure, I can add slightly more info; since an excerpt is more than just a text string, it seemed clearly implied to me. But I'll expand on the comment (but make `ruff` happy)

laurejt: To make this more explicit, I made the Excerpt a `TypedDict` which makes it more explicit what these structures are and also makes typing hints more accurate.

Issue #170, opened 2025-02-24
update process adjudication data script to use excerpt object and add excerpt id


mnaydan:
- keep output file format as csv
- only implement excerpt id construction if there is a *single* detection method
- make a quick look-up-table from detection methods to shorthand ids

Issue #169, opened 2025-02-24
edit existing excerpt object to add excerpt id property


mnaydan:
use cached property field

laurejt:
Additional decision made this morning (2/26)

- [ ] Excerpt objects are immutable
- [ ] Excerpt IDs can be optionally passed in during initialization, but will be validated during post initialization
- [ ] Add validation for detection method names

Issue #146, opened 2025-02-10
update Passim script to map to data class


mnaydan:

laurejt:
Note: This might impact assumptions made by the evaluation code.

PR #171, opened 2025-02-24
Feature/feature/add excerpt ids

laurejt:
ref: #169 #170 #146 

This script adds a bunch of additional functionality to `Excerpt` objects:
- excerpt id creation
- csv format support
- whitespace stripping logic

It also:
- updates the annotation script `processing_adjudication_data.py` to use `Excerpt` objects; and
- updates the passim script `get_passim_excerpts.py` to use `LabeledExcerpt` objects.

### DIFF
--- a/src/corppa/poetry_detection/annotation/process_adjudication_data.py
+++ b/src/corppa/poetry_detection/annotation/process_adjudication_data.py
@@ -134,7 +134,6 @@ def process_adjudication_data(
             page_data = process_page_annotation(page_anno)
 
             # Write excerpt-level data
-            print(page_data)
             for excerpt in page_data["excerpts"]:
                 csv_writer.writerow(excerpt.to_csv())
 
--- a/src/corppa/poetry_detection/annotation/process_adjudication_data.py
+++ b/src/corppa/poetry_detection/annotation/process_adjudication_data.py
@@ -24,52 +24,24 @@ import argparse
 import csv
 import pathlib
 import sys
-from collections.abc import Generator
-from typing import Any, TypedDict
+from collections.abc import Iterable
+from typing import Any
 
 import orjsonl
 from tqdm import tqdm
 from xopen import xopen
 
-
-class Excerpt(TypedDict):
-    """
-    Basic excerpt object, primarily used for type checking
-    """
-
-    start: int
-    end: int
-    text: str
-
-
-def clean_excerpt(span: Excerpt) -> Excerpt:
-    """
-    Clean excerpt so that any leading and trailing whitespace is removed. This involves
-    updating the excerpt text itself as well as its start and end indices.
-    """
-    updated_span = span.copy()
-    # Remove any leading whitespace
-    ldiff = len(span["text"]) - len(span["text"].lstrip())
-    if ldiff:
-        updated_span["start"] = span["start"] + ldiff
-        updated_span["text"] = updated_span["text"][ldiff:]
-    # Remove any trailing whitespace
-    rdiff = len(span["text"]) - len(span["text"].rstrip())
-    if rdiff:
-        updated_span["end"] = span["end"] - rdiff
-        updated_span["text"] = updated_span["text"][:-rdiff]
-    return updated_span
+from corppa.poetry_detection.core import Excerpt
 
 
 def get_excerpts(page_annotation: dict[str, Any]) -> list[Excerpt]:
     """
     Extract excerpts from page-level annotation. Excerpts have the following
-    fields:
         * start: character-level starting index
         * end: character-level end index (Pythonic, exclusive)
         * text: text of page excerpt
 
-    Note: Currently ignoring span labels, since there's only one for the
+    Note: This ignores span labels since there's only one for the
           poetry detection task.
     """
     excerpts = []
@@ -78,12 +50,14 @@ def get_excerpts(page_annotation: dict[str, Any]) -> list[Excerpt]:
     if "spans" not in page_annotation:
         raise ValueError("Page annotation missing 'spans' field")
     for span in page_annotation["spans"]:
-        excerpt: Excerpt = {
-            "start": span["start"],
-            "end": span["end"],
-            "text": page_text[span["start"] : span["end"]],
-        }
-        excerpts.append(clean_excerpt(excerpt))
+        excerpt = Excerpt(
+            page_id=page_annotation["id"],
+            ppa_span_start=span["start"],
+            ppa_span_end=span["end"],
+            ppa_span_text=page_text[span["start"] : span["end"]],
+            detection_methods={"adjudication"},
+        )
+        excerpts.append(excerpt.strip_whitespace())
     return excerpts
 
 
@@ -110,23 +84,22 @@ def process_page_annotation(page_annotation) -> dict[str, Any]:
     return page_data
 
 
-def get_excerpt_entries(page_data: dict[str, Any]) -> Generator[dict[str, Any]]:
+def simplify_excerpts(excerpts: Iterable[Excerpt]) -> list[dict[str, Any]]:
     """
-    Generate excerpt entries data from the processed page produced by
-    `process_page_annotation`.
+    Converts excerpts into a simplified form with the following fields:
+        * start = starting index of PPA span
+        * end = ending index of PPA span
+        * text = text of PPA span
     """
-    for excerpt in page_data["excerpts"]:
-        entry = {
-            "page_id": page_data["page_id"],
-            "work_id": page_data["work_id"],
-            "work_title": page_data["work_title"],
-            "work_author": page_data["work_author"],
-            "work_year": page_data["work_year"],
-            "start": excerpt["start"],
-            "end": excerpt["end"],
-            "text": excerpt["text"],
+    simplified_excerpts = []
+    for excerpt in excerpts:
+        simple_excerpt = {
+            "start": excerpt.ppa_span_start,
+            "end": excerpt.ppa_span_end,
+            "text": excerpt.ppa_span_text,
         }
-        yield entry
+        simplified_excerpts.append(simple_excerpt)
+    return simplified_excerpts
 
 
 def process_adjudication_data(
@@ -147,24 +120,29 @@ def process_adjudication_data(
     )
     csv_fieldnames = [
         "page_id",
-        "work_id",
-        "work_title",
-        "work_author",
-        "work_year",
-        "start",
-        "end",
-        "text",
+        "excerpt_id",
+        "ppa_span_start",
+        "ppa_span_end",
+        "ppa_span_text",
+        "detection_methods",
+        "notes",
     ]
     with open(output_excerpts, mode="w", newline="") as csvfile:
         csv_writer = csv.DictWriter(csvfile, fieldnames=csv_fieldnames)
         csv_writer.writeheader()
         for page_anno in progress_annos:
-            # Get & save page data
             page_data = process_page_annotation(page_anno)
-            orjsonl.append(output_pages, page_data)
 
-            for row in get_excerpt_entries(page_data):
-                csv_writer.writerow(row)
+            # Write excerpt-level data
+            print(page_data)
+            for excerpt in page_data["excerpts"]:
+                csv_writer.writerow(excerpt.to_csv())
+
+            # Simplify excerpts
+            # TODO: This is needed for compatiblity with existing evaluation code
+            page_data["excerpts"] = simplify_excerpts(page_data["excerpts"])
+            # Write page-level data
+            orjsonl.append(output_pages, page_data)
 
 
 def main():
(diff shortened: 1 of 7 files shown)


rlskoeser:
> @rlskoeser Should the new excerpt ID be included in the output of `to_dict`?

Yes! I think we'll want to include it in any form of excerpt we serialize (csv, json, etc)

laurejt:
> > @rlskoeser Should the new excerpt ID be included in the output of `to_dict`?
> 
> Yes! I think we'll want to include it in any form of excerpt we serialize (csv, json, etc)

Alright, then I'll go change the implementation so that IDs can be nullable; it's the only obvious work around for excerpts where the ID would be undefined.

src/corppa/poetry_detection/core.py
```diff
@@ -112,6 +135,38 @@ def to_dict(self) -> dict[str, Any]:
...
+
+    def strip_whitespace(self) -> "Excerpt":
+        """
+        Returns a copy of the excerpt such that leading and trailing whitespace
+        have been removed from the text. In addition to the text, the start
+        and end indices will also be updated.
+        """
+        ldiff = len(self.ppa_span_text) - len(self.ppa_span_text.lstrip())
+        rdiff = len(self.ppa_span_text) - len(self.ppa_span_text.rstrip())
+        return replace(
```
rlskoeser, line 163: `replace` -  good to know about this method! (had to look it up)

src/corppa/poetry_detection/core.py
```diff
@@ -112,6 +135,38 @@ def to_dict(self) -> dict[str, Any]:
...
+                csv_dict[key] = value
+        return csv_dict
+
+    def strip_whitespace(self) -> "Excerpt":
+        """
+        Returns a copy of the excerpt such that leading and trailing whitespace
+        have been removed from the text. In addition to the text, the start
+        and end indices will also be updated.
+        """
+        ldiff = len(self.ppa_span_text) - len(self.ppa_span_text.lstrip())
```
rlskoeser, line 161: If text length is something we need elsewhere we might consider adding a property

src/corppa/poetry_detection/core.py
```diff
@@ -112,6 +135,38 @@ def to_dict(self) -> dict[str, Any]:
...
+        rdiff = len(self.ppa_span_text) - len(self.ppa_span_text.rstrip())
+        return replace(
+            self,
+            ppa_span_start=self.ppa_span_start + ldiff,
+            ppa_span_end=self.ppa_span_end - rdiff,
+            ppa_span_text=self.ppa_span_text.strip(),
```
rlskoeser, line 167: I wondered about using the indices you already computed to substring the span, but this is probably more readable.

src/corppa/poetry_detection/core.py
```diff
@@ -129,7 +184,12 @@ class LabeledExcerpt(Excerpt):
     # Identification methods
     identification_methods: set[str]
 
+    # TODO: Should labeled excerpts have their own distinct IDs?
+    #       Right now, IDs are inherited from Excerpt class
```
rlskoeser, line 188: I think inheriting the id is what we want. Do you expect/imagine any difference in behavior?

src/corppa/poetry_detection/core.py
```diff
@@ -2,9 +2,17 @@
 Custom data type for poetry excerpts identified with the text of PPA pages.
 """
 
-from dataclasses import asdict, dataclass
+from dataclasses import asdict, dataclass, field, replace
 from typing import Any, Optional
 
+# Would upper casing be better?
+DETECTION_PFX_TABLE = {
```
rlskoeser, line 9: This might be a good use case for an enum in the future. This is fine, though.

rlskoeser, line 9: I would rename to just DETECTION_METHODS or something like that, since it's serving both purposes (list of valid methods as well as the id prefix).

src/corppa/poetry_detection/core.py
```diff
@@ -85,16 +93,31 @@ class Excerpt:
...
+        if len(self.detection_methods) == 1:
+            [detect_name] = self.detection_methods
+            if detect_name not in DETECTION_PFX_TABLE:
+                raise ValueError(f"Unsupported detection method '{detect_name}'")
+            detect_pfx = DETECTION_PFX_TABLE[detect_name]
+            self.excerpt_id = f"{detect_pfx}@{self.ppa_span_start}:{self.ppa_span_end}"
+        else:
+            # TODO: Should all detection methods be validated?
+            # TODO: Is this the tag we want for multiple detection methods?
+            self.excerpt_id = f"mult@{self.ppa_span_start}:{self.ppa_span_end}"
```
rlskoeser, line 120: For the detection method check, it's more pythonic to just try to access and then catch if it's not found.

I suggest adjusting the logic so the id construction is done once.

```suggestion
        if len(self.detection_methods) == 1:
            [detect_name] = self.detection_methods
            try:
                id_prefix = DETECTION_PFX_TABLE[detect_name]
            except KeyError:
                raise ValueError(f"Unsupported detection method '{detect_name}'")
            
            self.excerpt_id = f"{detect_pfx}@{self.ppa_span_start}:{self.ppa_span_end}"
        else:
            id_prefix = "m" 
           
        self.excerpt_id = f"{id_prefix}@{self.ppa_span_start}:{self.ppa_span_end}"        
```

rlskoeser: Regarding your TODOs:
- yes, probably should validate detection methods but it doesn't need to be implemented now, it's not what you're working on (you can leave a todo if you want)
- can we reserve `m` for the prefix of combined records? m for multi/merge

src/corppa/poetry_detection/core.py
```diff
@@ -112,6 +135,38 @@ def to_dict(self) -> dict[str, Any]:
...
+        csv_dict = {}
+        for key, value in asdict(self).items():
+            if value is None:
+                # Set unset fields as empty strings
+                csv_dict[key] = ""
+            elif type(value) is set:
+                # Convert sets to comma-separated lists
+                csv_dict[key] = ", ".join(value)
+            else:
+                csv_dict[key] = value
```
rlskoeser, line 152: I was originally wondering why we needed this, but after using the current code I see the need for flattening lists!

I don't think the None / empty string matters for CSV.

For both this method and `to_dict` we should probably be taking advantage of the `dict_factory` parameter. (But no need to do that now)

src/corppa/poetry_detection/core.py
rlskoeser: ooh but when we do the detection method validation we should use set logic

src/corppa/poetry_detection/core.py
```diff
@@ -112,6 +135,38 @@ def to_dict(self) -> dict[str, Any]:
...
+        return csv_dict
+
+    def strip_whitespace(self) -> "Excerpt":
+        """
+        Returns a copy of the excerpt such that leading and trailing whitespace
+        have been removed from the text. In addition to the text, the start
+        and end indices will also be updated.
```
rlskoeser, line 159: Suggested edit to make it a bit more readable

```suggestion
        Return a copy of this excerpt with any leading and trailing whitespace
        removed from the text and start and end indices updated to match any changes.
```

rlskoeser (approved):
Looks great. I like the way you refactored the whitespace cleaning method.

I have some comments and a few suggestions for minor improvements, but nothing that really needs to be changed now. I think it's ready to merge when you're satisfied with it.

src/corppa/poetry_detection/core.py
laurejt: Agreed, although I have a feeling this should actually be incorporated into the `Span` object itself (then again incorporating `Span` objects into the `Excerpt` objects has been out of scope for now)

src/corppa/poetry_detection/annotation/process_adjudication_data.py
```diff
@@ -147,24 +120,29 @@ def process_adjudication_data(
...
         csv_writer.writeheader()
         for page_anno in progress_annos:
-            # Get & save page data
             page_data = process_page_annotation(page_anno)
-            orjsonl.append(output_pages, page_data)
 
-            for row in get_excerpt_entries(page_data):
-                csv_writer.writerow(row)
+            # Write excerpt-level data
+            print(page_data)
```
rlskoeser, line 137: In case you forgot, you left a print statement in (noticed when I ran the script just now).

src/corppa/poetry_detection/core.py
laurejt: One related question, is whether an `Excerpt` object should be immutable. The way excerpt ids are constructed, suggest it is; but I wasn't sure we explicitly wanted that behavior.

src/corppa/poetry_detection/core.py
```diff
@@ -85,16 +93,31 @@ class Excerpt:
...
+            raise ValueError("Must specify at least one detection method")
+
+        # Set excerpt id
+        # TODO: Define behavior for multiple detection methods
+        if len(self.detection_methods) == 1:
+            [detect_name] = self.detection_methods
+            if detect_name not in DETECTION_PFX_TABLE:
+                raise ValueError(f"Unsupported detection method '{detect_name}'")
+            detect_pfx = DETECTION_PFX_TABLE[detect_name]
+            self.excerpt_id = f"{detect_pfx}@{self.ppa_span_start}:{self.ppa_span_end}"
```
rlskoeser, line 116: I meant to ask: why did you decide to construct the id instead of making it a property?

src/corppa/poetry_detection/core.py
laurejt: One reason to keep it separate is if the span indices don't actually match the text itself...

Although one issue with this method is we could attempt to construct an invalid `Excerpt` (i.e., `ppa_span_end <= ppa_span_start`)

src/corppa/poetry_detection/core.py
laurejt: I mostly wanted to flag that there is no guarantee of these being unique for labeled excerpts. Also there's a related strong worry about what mutation we allow for labeled excerpts... can a span's indices and/or text be modified?

src/corppa/poetry_detection/core.py
laurejt: No, multi/merge cannot be m because "m" is already in use for "manual"

src/corppa/poetry_detection/core.py
rlskoeser: I wouldn't expect them to be... but you might be convincing me that the id should be a dynamic property

src/corppa/poetry_detection/core.py
laurejt: You're probably write on unset fields front; I did it so I didn't have to think about the necessary settings needed for writing the CSV to file. Since fields could be missing / inconsistent between excerpts.

src/corppa/poetry_detection/core.py
laurejt: Yesterday, I believe we agreed that we wanted the `Excerpt` class to define the form of IDs. A quick issue that arose when it was a simple `cached_property` was that if we didn't call the cached_property than it wouldn't be class variable. Which would cause trouble for `to_dict`. Instead of adding complexity to the `to_dict` function, it seemed easier to make excerpt ids a constructed class variable set in post initialization.

rlskoeser: Ok, thanks for explaining. That makes sense.

src/corppa/poetry_detection/core.py
laurejt: ooh, c for combined / combination is a nice alternative.

maybe there's a better alternative to manual; but I wasn't sure. it's a catch-all for non-adjudication manual entries, but maybe we can make a quick decision tomorrow morning.

src/corppa/poetry_detection/core.py
rlskoeser: Might be worth documenting this briefly in a comment somewhere for our future selves. (defined as a property for simplicity of export)

New question: what happens when we initialize an excerpt that already has an id? Should we validate that it matches the constructed id?

src/corppa/poetry_detection/core.py
laurejt: I think there's value in having a rigid detected (unlabeled) excerpt ID, it just seems like this might differ from the labeled excerpt scenario; I think there are a bunch of options but it's just figuring out what best aligns with our needs. No need to build a bunch of support for infrastructure we don't actually need.

src/corppa/poetry_detection/core.py
laurejt: Setting to "c" for combined.

laurejt:
@rlskoeser It looks like I mispoke about being able to pass in unnecessary keyword arguments in initialization. To circumvent this issue, I created a `from_dict` method that constructs an `Excerpt` (or `LabeledExcerpt`) from a dict of the form created by `to_dict` or `to_csv`.  Can you please review this new functionality before I merge this branch?

rlskoeser (approved):
All the revisions look good to me. Anything you want me to test before you merge? My inclination is to test by way of starting to use this in my code after you merge it.

PR #172, opened 2025-02-27
Revise refmatcha script to use Excerpt and LabeledExcerpt objects

rlskoeser:
towards #150 

changes in this script:
- renamed refmatcha script to use the new name, and configured to be installed when the package is installed
- input csv is now expected to be in `Excerpt` format
- output csv is now created using `LabeledExcerpt`
  - updated regex matching logic to return reference span indices and text 
- added & tested a new method to `Excerpt` to make it easy to get a list of fieldnames

Current output for adjucation round 2 data: [allround2_excerpts_matched.csv](https://github.com/user-attachments/files/19013944/allround2_excerpts_matched.csv)


src/corppa/poetry_detection/refmatcha.py
laurejt, line 1: It'd be nice to have method level descriptions, and typing hints if possible. It's not always obvious what the various inputs / outputs are.

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -309,7 +316,7 @@ def fuzzy_partial_ratio(series, search_text):
     return scores[0]
 
 
-def find_reference_poem(ref_df, input_row, meta_df):
+def find_reference_poem(input_row, ref_df, meta_df):
```
laurejt, line 319: What are the inputs of this function? What does this function output?

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -325,16 +332,68 @@ def find_reference_poem(ref_df, input_row, meta_df):
...
             # do a case-insensitive search
-            result = ref_df.filter(
-                pl.col("search_text").str.contains(f"(?i){search_text}")
+            re_search = f"(?i){search_text}"
+            re_extract = f"(?i)^(?<preceding_text>.*)(?<ref_span_text>{search_text})"
+            result = (
+                # filter to rows that match the regex search
+                ref_df.filter(pl.col("search_text").str.contains(re_search))
+                # for those that match, extract the search text AND
+                # the all the text that comes before it
```
laurejt, line 341:
```suggestion
                # all of the text that comes before it
```

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -325,16 +332,68 @@ def find_reference_poem(ref_df, input_row, meta_df):
...
+                .with_columns(
+                    captures=pl.col("search_text").str.extract_groups(re_extract)
+                )
+                .unnest("captures")
+                # calculate start based on character length of text before the match
+                # NOTE: can't use str.find because it returns byte offset instead of char offset
+                .with_columns(ref_span_start=pl.col("preceding_text").str.len_chars())
+                # calculate span end based on span start and length of matching text
```
laurejt, line 349: This is a estimate / best effort, right? In other words, there's no guarantee that the matching reference text would be the same length as the starting excerpt, right?

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -325,16 +332,68 @@ def find_reference_poem(ref_df, input_row, meta_df):
...
+                )
+
+            # then extract full reference text for adjusted indices
+            result = result.with_columns(
+                ref_span_text=pl.col("search_text").str.slice(
+                    pl.col("ref_span_start"), search_text_length
+                )
+            )
+
         # otherwise, see if the results are useful
```
laurejt, line 397: Is this a residual comment?

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -456,16 +519,28 @@ def main(input_file):
...
     # split out text to isolate first and last linesfind based on
     input_df = input_df.with_columns(
-        first_line=pl.col("text").str.strip_chars().str.split("\n").list.first(),
-        last_line=pl.col("text").str.strip_chars().str.split("\n").list.last(),
+        first_line=pl.col("ppa_span_text")
+        .str.strip_chars()
+        .str.split("\n")
+        .list.first(),
+        last_line=pl.col("ppa_span_text").str.strip_chars().str.split("\n").list.last(),
     )
```
laurejt, line 544: Is there any way this could be separated into a function that's called for both the search and reference text?

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -476,63 +551,80 @@ def main(input_file):
...
-        "match_poem_id",
-        "match_author",
-        "match_title",
-        "match_notes",
-    ]
 
     with output_file.open("w", encoding="utf-8") as outfile:
         # add byte-order-mark to indicate unicode
-        outfile.write(codecs.BOM_UTF8)
+        outfile.write(codecs.BOM_UTF8.decode())
```
laurejt, line 568: Why are you writing including the BOM? This is generally unnecessary.

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -476,63 +551,80 @@ def main(input_file):
...
-                match_poem["match_notes"] = match_poem.pop("notes")
-                row.update(match_poem)
-
-                # update the tally of rows we found matches for
+                # the dataframe row now includes search text fields;
+                # filter to just excerpt fields
+                excerpt_fields = {
+                    k: v for k, v in row.items() if k in Excerpt.fieldnames()
+                }
+                notes = excerpt_fields.pop("notes") or ""
```
laurejt, line 587:
```suggestion
                notes = excerpt_fields.pop("notes", "")
```
`pop` can take a default value as input

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -476,63 +551,80 @@ def main(input_file):
...
+                    poem_id=match_poem["id"],
+                    ref_corpus=match_poem["source"],
+                    ref_span_start=match_poem["ref_span_start"],
+                    ref_span_end=match_poem["ref_span_end"],
+                    ref_span_text=match_poem["ref_span_text"],
+                    identification_methods={"refmatcha"},
+                    notes=notes,
+                    # TODO: use new from_dict method ?
+                    **excerpt_fields,
+                )
```
laurejt, line 603: It seems like we probably want a new constructor for `LabeledExcerpt` one that involves labeling an existing `Excerpt`.

laurejt: As an aside, because Python isn't statically typed we might want to be more proactive in the enforcement of sets for `detection_methods` and `identification_methods`, not quite sure how this would work with dataclasses but it seems plausible that we could accept any kind of Iterable as input arguments and convert to sets in post-initialization

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -476,63 +551,80 @@ def main(input_file):
...
                 match_found += 1
-            else:
-                row["match_count"] = 0
-            # write row out with input and any match information found;
-            # filter out any fields not in the list of csv field names
-            csvwriter.writerow(row)
+                # TODO: use new to_csv method
+                csvwriter.writerow(excerpt.to_dict())
```
laurejt, line 606:
```suggestion
                csvwriter.writerow(excerpt.to_csv())
```

laurejt: `to_dict` is for better for JSONL-style output rather than CSV

src/corppa/poetry_detection/refmatcha.py
rlskoeser: If we match on the full text of the excerpt, then the reference span and text should be accurate. In any other case (i.e., match found based on first line or last line) it will be best effort. (Not sure yet about fuzzy matches)

src/corppa/poetry_detection/refmatcha.py
rlskoeser: This tells Excel to open as unicode, so you can open directly instead of having to do an open/import and explicitly set the encoding. I tested after our last exchange about this, specifying the encoding when you open the file does NOT write the byte-order mark to the file.

src/corppa/poetry_detection/refmatcha.py
rlskoeser: yes, but it's possible to have an actual/non-default value of `None` and I need a string here (I'll add comments to clarify)

src/corppa/poetry_detection/refmatcha.py
rlskoeser: @laurejt I wondered about the most efficient way to create a LabeledExcerpt from an Excerpt - would love to hear what you're thinking here because that would be pretty convenient.

Also love the idea of allowing any iterable for the methods and converting to sets in in post-init, that would be useful.

src/corppa/poetry_detection/refmatcha.py
laurejt: Fair enough, I had been assuming the Excerpt structure and so if it was None it wouldn't exist; but that's not true in this case I guess?

src/corppa/poetry_detection/refmatcha.py
laurejt: Ah, this is a Windows thing. Since BOM isn't inherently a UTF-8 thing; is there any special handling required if the mark is not guaranteed to exist?

rlskoeser:
Whew! 😮‍💨  Finally got test coverage over our threshold - and found and fixed multiple problems along the way.

This updates refmatcha to use excerpt and labeled excerpt objects. It disables the fuzzy matching logic, improves white space handling, and adds unit test and documentation.

@laurejt Please be aware: this update does NOT include an update to the new version of the Chadwyck-Healey metadata and filtered text files; I will handle that as a separate step per our decision on the checklist on #150 as a separate PR after this one.  It should be fairly easy now that you've done the filtering work for me, but the metadata file is slightly different so it will require a few changes.

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -1,18 +1,27 @@
...
 """
 
-🎶🍵 matcha matcha poem / This script is gon / na find your poems / matcha matcha poem 🎶🍵
+*🎶🍵 matcha matcha poem / This script is gon / na find your poems / matcha matcha poem 🎶🍵*
 
-refmatcha is a script to identify poem excerpts by matching against a local
-collection of reference poems.  It takes in a CSV of excerpts
-and outputs a CSV of labeled excerpts for those excerpts it is able to identify.
+refmatcha identifies poem excerpts by matching against a local collection of
+reference poems.  It takes in a CSV of excerpts and outputs a CSV of
```
laurejt, line 7:
```suggestion
reference poems.  It takes in a CSV of detected (i.e. unidentified) excerpts and outputs a CSV of
```
An alternative could be explicitly mention a CSV of poetry_detection.core.Excerpt data

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -1,18 +1,27 @@
...
+By default, the output file is created in the same directory as the input
+with the same name plus `_matches`; i.e., given an input file `round1_excerpts.csv`,
+refmatcha will output identified excerpts to `round1_excerpts_matches.csv`.
+To override this, specify an output filename with `--output` or `-o`.
 
 Setup:
 
-Download and extract poetry-ref-data.tar.bz2 from /tigerdata/cdh/prosody/poetry-detection
+Download and extract `poetry-ref-data.tar.bz2` from `/tigerdata/cdh/prosody/poetry-detection`.
 You should extract it in the same directory where you plan to run this script.
```
laurejt, line 17: Is this tarball required to run refmatcha? How would I run refmatcha on a new reference corpus? Maybe this is out of scope, but it'd be useful to have a bit more of these implicit inputs documented.

Would it make sense to have an option to provide a reference corpus directory and corresponding reference ID as inputs? Totally understand if this functionality is out of scope.

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -53,6 +72,7 @@
 SOURCE_ID = {
     "Poetry Foundation": "poetry-foundation",
```
laurejt, line 73: Is it worth keeping this in, given the decision to not use poetry foundation?

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -326,127 +374,143 @@ def fuzzy_partial_ratio(series, search_text):
...
-                .with_columns(ref_span_start=pl.col("preceding_text").str.len_chars())
-                # calculate span end based on span start and length of matching text
-                .with_columns(
-                    ref_span_end=pl.col("ref_span_start").add(
-                        pl.col("ref_span_text").str.len_chars()
-                    )
+def identify_excerpt(
+    excerpt_row: dict, reference_df: pl.DataFrame, search_text: str = "text"
+) -> LabeledExcerpt | None:
+    """Given an unlabeled excerpt as a dict from a polars dataframe and a
```
laurejt, line 380:
```suggestion
    """Given an unlabeled excerpt as a dict from a polars DataFrame and a
```

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -326,127 +374,143 @@ def fuzzy_partial_ratio(series, search_text):
...
+    reference poetry data frame, attempt to identify the excerpt. Returns
```
laurejt, line 381:
```suggestion
    reference poetry dataframe, attempt to identify the excerpt. Returns
```

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -326,127 +374,143 @@ def fuzzy_partial_ratio(series, search_text):
...
+    identification and reference information if found."""
+    # can we use excerpt objects to simplify?
+    excerpt = Excerpt.from_dict(
+        {k: v for k, v in excerpt_row.items() if k in EXCERPT_FIELDS}
+    )
+    match_info = None
+    result = None
+    # preserve any notes on the incoming excerpt
+    # (is this what we want? notes might get duplicated if/when we merge...)
+    note_lines = [excerpt_row["notes"]] if excerpt.notes is not None else []
```
laurejt, line 392: It looks like there's duplicating behavior here with the merge script. How much of an issue does this cause us? If it's enough of a problem, we should resolve this issue in the near future.

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -326,127 +374,143 @@ def fuzzy_partial_ratio(series, search_text):
...
+            # all of the text that comes before it
+            .with_columns(captures=pl.col("search_text").str.extract_groups(re_extract))
+            .unnest("captures")
+            # calculate start based on character length of text before the match
+            # NOTE: can't use str.find because it returns byte offset instead of char offset
+            .with_columns(ref_span_start=pl.col("preceding_text").str.len_chars())
+            # calculate span end based on span start and length of matching text
+            .with_columns(
+                ref_span_end=pl.col("ref_span_start").add(
+                    pl.col("ref_span_text").str.len_chars()
```
laurejt, line 422: Does this text correspond to an excerpt from the untransformed PPA text corpus?

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -326,127 +374,143 @@ def fuzzy_partial_ratio(series, search_text):
...
-            # match poem includes id, author, title
-            match_poem = result.to_dicts()[0]
+            match_df = result
+            id_note = f"single match on {search_field_label}"
+        elif num_matches < 10:
+            # if there are a small number of matches, check if we have
+            # duplicates across corpora and they agree
+            match_df, reason = multiple_matches(result)
+            if match_df is not None:
+                id_note = f"{num_matches} matches on {search_field_label}: {reason}"
```
laurejt, line 447: Is it worth making a note for the final case (i.e., `num_matches >= 10`)?

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -326,127 +374,143 @@ def fuzzy_partial_ratio(series, search_text):
...
-    return
+# NOTE: disabling code coverage since this code is skipped for now
+def _find_reference_poem_OLD(input_row, ref_df, meta_df):  # pragma: no cover
+    # NOTE: this is the old version of the identification method
+    # which is now replaced by identify_excerpt
+    # fuzzy matching will be updated and restored later
 
-    # NOTE: could make configurable to skip fuzzy searches,
-    # or maybe truncate large excepts before running fuzzy search
+    # TODO: add argparse to make it configurable whether to try fuzzy matchingd
```
laurejt, line 512:
```suggestion
    # TODO: add argparse to make it configurable whether to try fuzzy matching
```

laurejt (approved):
Overall seems fine, two points worth documenting and possibly addressing are (1) the assumptions logic going on with the implicit reference corpora and (2) where the `ppa_span_text` field corresponds to the "definitive" excerpts from the PPA text corpus.

I don't think either these are must fixes, but it's worth keeping track of these things. The more noisy the `ppa_span_text` field is, we may want to revisit how merging logic works.

src/corppa/poetry_detection/refmatcha.py
rlskoeser: I just checked in the current merged data - it isn't a problem for the current set of excerpts because refmatcha was not run on any excerpts that had notes, so I'm not getting any duplication. We might in future with the eebo-xml content if we merge both versions of the excerpt (unidentified and refmatcha labeled), but I'm inclined to leave it as is for now until we know if it's an actual problem for our data.

Maybe we can keep in mind when work on formalizing and documenting our pipeline.

src/corppa/poetry_detection/refmatcha.py
rlskoeser: interesting question... right now we only return an excerpt if we are able to identify a match, since output assumes a labeled excerpt. I think you're proposing an unlabeled excerpt with a note and this needs to be out of scope for now (but I could see it being valuable).

src/corppa/poetry_detection/refmatcha.py
rlskoeser: I'll try to figure out a way to make the reference corpora slightly more modular when I revisit the chadwyck- healey part (also possible overlap with our need to create a filtered/combined poetry metadata csv for the dataset).

Right now I have it structured so the text is all compiled at once and then the metadata is all compiled at once - I can see it would be better if it was sort of "provider specific" so it's easier to add a new corpus.

src/corppa/poetry_detection/refmatcha.py
laurejt: Ah, right; this is the fail case. Definitely seems out of scope for now.

src/corppa/poetry_detection/refmatcha.py
rlskoeser: This script doesn't modify the incoming ppa span text at all.

The reference text it pulls in right now is the modified version for searching — it doesn't always match exactly (esp. when match is based on first or last line), but for the ones I've looked at it seems close enough to be useful.  We discussed a while back to prioritize the ppa text content and allow the dataset to be looser with the reference text, which I was grateful for here. Maybe in future we could use your alignment logic to get a better version of the reference text span.

src/corppa/poetry_detection/refmatcha.py
rlskoeser: OH! but this is dead code - this is when I was outputting a csv for review of the matching, this note doesn't go anywhere anymore. I'll clean it up.

Issue #156, opened 2025-02-10
write script that combines detected spans with poem identifications


mnaydan:
work with both manual identifications (#148 ) and revise output of refmatcha (#150 )

script input:
- 1 CSV of detected excerpts 
- one or more CSVs of identified excerpts

script output:
- JSONL file of labeled excerpts

if multiple ids for the same excerpt, merge; if poem identifications for the same span don't match, log a warning and output two versions of the excerpt (keeps the same excerpt id)

## revisions after review
- [x] don't collapse different poem ids
- [x] make sure method handling is order agnostic
   - [x] fix set/multival delimiter; make a constant
- [x] separate pr for excerpt revisions
   - [x] get field types from excerpt objects
   - [x] test new from_excerpt method
- [x] decide how to handle columns when opening files (minimum required; report whether we have labeled/unlabeled coming in)
- [x] document assumption of no duplicates within single input
- [x] strip notes white space before combining

rlskoeser:
Script is merged into develop; initial exploration in the notebook suggests that somewhere this is introducing duplicates - we're getting unlabeled and labeled versions of the same excerpt, which isn't what I expected - but we'll close this and deal with that later.

PR #174, opened 2025-02-28
Merge unlabeled and labeled excerpts 

rlskoeser:
Preliminary version of script for #156  - looking for feedback on the general approach and agreement on what kind of merging in scope for now (basically only simple cases: unlabeled + labeled, labeled + labeled where ids match and ref span doesn't conflict).

Handles the simple cases of merging unlabeled excerpt with labeled excerpt (e.g. annotation and manual ids, or annotation + refmatcha output), and can merge multiple ids for the same excerpt for simple cases (ref spans match or one ref span unset).

I need to settle on treating methods as strings or lists; I think strings would be easier for combination. Also, we should probably use a delimiter _other_ than comma in our `to_csv` method when combining sets (sorry I didn't flag this when I reviewed).  This problem is what's causing one of the tests to fail.

Here's what the script looks like when run with round 2 annotation data:
```console
% ./src/corppa/poetry_detection/merge_excerpts.py finish-line/adjudication/adjudicated_excerpts_sample.csv finish-line/adjudication/round2_excerpts_matched.csv finish-line/adjudication/round2_excerpts.csv -o round2_combined.csv
Loaded 1,291 excerpts from 3 files.
1,002 total excerpts after merging; 385 labeled excerpts.
Total by detection method:
        ['adjudication']: 172
        adjudication: 830
        manual: 175
        refmatcha: 172
        manual,refmatcha: 38
```

Here's the resulting output:
[round2_combined.csv](https://github.com/user-attachments/files/19029076/round2_combined.csv)


laurejt:
> Preliminary version of script for #156 - looking for feedback on the general approach and agreement on what kind of merging in scope for now (basically only simple cases: unlabeled + labeled, labeled + labeled where ids match and ref span doesn't conflict).
> 
> Handles the simple cases of merging unlabeled excerpt with labeled excerpt (e.g. annotation and manual ids, or annotation + refmatcha output), and can merge multiple ids for the same excerpt for simple cases (ref spans match or one ref span unset).
> 
> I need to settle on treating methods as strings or lists; I think strings would be easier for combination. Also, we should probably use a delimiter _other_ than comma in our `to_csv` method when combining sets (sorry I didn't flag this when I reviewed). This problem is what's causing one of the tests to fail.
> 
> Here's what the script looks like when run with round 2 annotation data:
> 
> ```
> % ./src/corppa/poetry_detection/merge_excerpts.py finish-line/adjudication/adjudicated_excerpts_sample.csv finish-line/adjudication/round2_excerpts_matched.csv finish-line/adjudication/round2_excerpts.csv -o round2_combined.csv
> Loaded 1,291 excerpts from 3 files.
> 1,002 total excerpts after merging; 385 labeled excerpts.
> Total by detection method:
>         ['adjudication']: 172
>         adjudication: 830
> Total by identification method:
>         manual: 175
>         refmatcha: 172
>         manual,refmatcha: 38
> ```
> 
> Here's the resulting output: [round2_combined.csv](https://github.com/user-attachments/files/19029076/round2_combined.csv)

What do you mean by strings vs. lists? These should be set comparisons...

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,256 @@
...
+    "ref_span_text": str,
+    "ref_span_start": int,
+    "identification_methods": str,
+}
+
+
+def excerpts_df(input_file: pathlib.Path) -> pl.DataFrame:
+    """Load the specified input file as a Polars dataframe,
+    with column names based on fields in
+    :class:`~corppa.poetry_detection.core.LaebledExcerpt`."""
```
laurejt, line 30:
```suggestion
    :class:`~corppa.poetry_detection.core.LabeledExcerpt`."""
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,256 @@
...
+
+FIELD_TYPES = {
+    "notes": str,
+    "ref_span_end": int,
+    "poem_id": str,
+    "ref_corpus": str,
+    "ref_span_text": str,
+    "ref_span_start": int,
+    "identification_methods": str,
+}
```
laurejt, line 24: Why aren't you grabbing these directly from `LabeledExcerpt`?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,256 @@
...
+    # TODO: requiring all fields means adjudication data poem ids are ignored
+    # what is the minimum set we need to check for?
+    return pl.read_csv(input_file)  # , columns=LABELED_EXCERPT_FIELDS)
+
+    try:
+        return pl.read_csv(input_file, columns=LABELED_EXCERPT_FIELDS)
+    except pl.exceptions.ColumnNotFoundError as err:
+        print(err)
+        # if label fields are missing, load as unlabeled excerpts
+        return pl.read_csv(input_file, columns=EXCERPT_FIELDS)
```
laurejt, line 44: Seems like this is a case where the `columns` should be specified as an input argument to `excerpts_df`

src/corppa/poetry_detection/core.py
```diff
@@ -255,5 +258,15 @@ def from_dict(d: dict) -> "LabeledExcerpt":
...
+                )
+
         return LabeledExcerpt(**input_args)
+
+    @staticmethod
+    def from_excerpt(ex: Excerpt, **kwargs: dict) -> "LabeledExcerpt":
+        excerpt_info = asdict(ex)
+        excerpt_info.update(kwargs)
+        excerpt_info.pop("excerpt_id")
+        return LabeledExcerpt(**excerpt_info)
```
laurejt, line 272: One thing I want to flag in the current implementation is that there is still some possibility of state mutation caused by set references. To avoid it in this specific instance would presumably require deepcopying (like in from_dict). I just realized that this deepcopying should be probably be taking place within post-initialization anyway.

We might also just be hitting the limits of dataclasses usefulness as we move further and further away from a basic container object.

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,256 @@
...
+    All excerpts from both dataframes are included in the resulting dataframe.
+    Excerpts are combined as follows:
+    - an unlabeled excerpt and a labeled excerpt for the same excerpt
+      will be combined
+    - if combined excerpts both have content in the notes, the notes text
+      will be combined
+    - multiple labeled excerpts for the same excerpt id are NOT combined
+    """
+
+    # smplest option is to do a LEFT join on page id and excerpt id
```
laurejt, line 82:
```suggestion
    # simplest option is to do a LEFT join on page id and excerpt id
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,256 @@
...
+    """Combine two Polars dataframes with excerpt or labeled excerpt data.
+    Excerpts are joined on the combination of page id and excerpt id.
+    All excerpts from both dataframes are included in the resulting dataframe.
+    Excerpts are combined as follows:
+    - an unlabeled excerpt and a labeled excerpt for the same excerpt
+      will be combined
+    - if combined excerpts both have content in the notes, the notes text
+      will be combined
+    - multiple labeled excerpts for the same excerpt id are NOT combined
+    """
```
laurejt, line 80: Does this assume that no merging is necessary within a single dataframe?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,256 @@
...
+
+    # smplest option is to do a LEFT join on page id and excerpt id
+    join_fields = ["page_id", "excerpt_id"]
+    # before joining, drop redundant fields that will be the same
+    # on any excerpt with matching page & excerpt id
+    other_join = other_df.drop(
+        "detection_methods", "ppa_span_start", "ppa_span_end", "ppa_span_text"
```
laurejt, line 87: Does this make assumptions about the ordering of the detection_methods field?

laurejt: Also, what if their notes field differed for some reason? (It's fine if this case isn't possible, but I'd like that to be more explicitly documented)

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,256 @@
...
+    )
+    merged = df.join(other_join, on=join_fields, how="left")
+
+    # NOTE: could also have right columns for poem_id_right, ref_corpus_right
+    # if they are the same, merge; otherwise... split into two rows
```
laurejt, line 92: This definitely seems like an issue if they don't have the same poem_id...

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,256 @@
...
+
+    # if notes_right is present, then we have notes coming from both sides
+    # of the join; combine the notes into a single notes field
+    if "notes_right" in merged.columns:
+        # update notes field by combining left and right notes with a newline,
+        # and then strip any outer newlines
+        merged = merged.with_columns(
+            notes=pl.col("notes")
```
laurejt, line 100: Should we strip any trailing whitespace before concatenation?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,256 @@
...
+        merged = merged.with_columns(
+            identification_methods=pl.when(
+                pl.col("identification_methods_right").is_null()
+            )
+            .then(pl.col("identification_methods"))
+            .when(pl.col("identification_methods").is_null())
+            .then(pl.col("identification_methods_right"))
+            .otherwise(
+                pl.col("identification_methods")
+                .add(",")  # TODO: we should probably use a different delimiter...
```
laurejt, line 116: It's a two-character delimiter `", "`. Moreover, since these represent sets, you need to worry about the ordering (direct string comparison is not inherently good)

laurejt (commented):
This is only a partial review, but I wanted to have these comments reach you faster than I'll finish this.

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: whoops, thank you

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: We don't have types directly accessible right now, so this was a shortcut - they should ideally be pulled from the class. I have work in progress to generate this from the class - please let me know if I'm missing something and we already have access!

src/corppa/poetry_detection/core.py
rlskoeser: ah, good point - should be easy enough to construct a test that triggers this behavior

do you think it's an immediate concern? are we likely to run into this in the short term?

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: yes -  the one exception is if the final merged set has excerpts with duplicate identifications that can be easily merged.  I'll update and document that assumption explicitly (I do think it's a reasonable assumption for our data at this point, but please LMK if there's something I'm missing)

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: yes, I just identified this as I was putting up for review; this is my highest priority to handle next

rlskoeser:
> What do you mean by strings vs. lists? The methods fields are sets, so it should just be set operations.

I'm working with polars dataframes, which doesn't have a set type. Your comment prompted me to do some looking, and it seems that the best match is [categorical or enum](https://docs.pola.rs/user-guide/expressions/categorical-data-and-enums/); but I can't find any multivalue examples, so we'd still need a list and this doesn't solve the ordering problem for us. I'll make a note to make sure the implementation is not dependent on order.

src/corppa/poetry_detection/merge_excerpts.py
laurejt: Those look like the right fields to me; that said, `identification_methods` should be a set (but maybe this is where the other PR comes into play)

src/corppa/poetry_detection/core.py
laurejt: I think we'll be okay in the short-term. Mostly just wanted to flag the issue.

src/corppa/poetry_detection/merge_excerpts.py
laurejt: I think you're right with whats going to be merged when it comes to unlabeled excerpts. The one case that comes to mind is regarding adjudication excerpts that actually contain multiple separate excerpts. I expect I'll be manually creating a few labeledexcerpts of that form, but I don't think this will impact this script.

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: correct - I've updated to use the new field_types method and found a solution for the set fields

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: I don't think it matters for this script; you should add a comment to the notes if you haven't already (I think you already did this?). If the identifications match they will get merged, otherwise they'll be two different entries. When you split out yours version, we can merge those in and they will be treated as separate excerpts since the span indices and ids will be different.

rlskoeser:
@laurejt I need to write a few more unit tests, but everything else has been addressed; this branch has been rebased on the excerpt refactor and depends on those changes.  The most important updates are handling merges on poem ids and consolidating identification methods. I've also added documentation at the top of the script explaining what it does, and tried to clarify the assumptions and limitations.

Here's the output and results from a recent run on round1 data:

```
./src/corppa/poetry_detection/merge_excerpts.py finish-line/adjudication/round1_excerpts.csv finish-line/adjudication/round1_excerpts_matched.csv finish-line/adjudication/adjudicated_excerpts_sample.csv -o round1_merged.csv
Loaded 2,537 excerpts from 3 files.
1,812 total excerpts after merging; 822 labeled excerpts.
Total by detection method:
        adjudication: 1,812
        refmatcha: 609
        manual; refmatcha: 58
        manual: 155
```

[round1_merged.csv](https://github.com/user-attachments/files/19076573/round1_merged.csv)

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,354 @@
...
+    if not right_df.is_empty():
+        # ensure field order and types match, then append the
+        # excerpts from the right dataframe to the end of the merged dataframe
+        merged = fix_columns(merged)
+        merged = merged.select(LABELED_EXCERPT_FIELDS).extend(fix_columns(right_df))
+
+    return merged
+
+
+def merge_duplicate_ids(df):
```
laurejt, line 209: Is it accurate to say that this method, merges labeled excerpts? Perhaps we could give this method a better name...

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,354 @@
...
+    # group by page id, excerpt id, and poem id to find repeated identificatins
+    for group, data in updated_df.group_by(["page_id", "excerpt_id", "poem_id"]):
+        # group is a tuple of values for page id, excerpt id, poem id
+        # data is a df of the grouped rows for this set
+
+        # sort so any empty values for optional fields are last,
+        # then fill values forward - i.e., treat nulls as duplicates
+        data = data.sort(
+            "ref_span_start", "ref_span_end", "ref_span_text", nulls_last=True
+        ).select(pl.all().forward_fill())
```
laurejt, line 227: So, what if the optional reference values (indicies or text) are only partially specified? Is it disagreement because the fields aren't exactly the same?

src/corppa/poetry_detection/core.py
```diff
@@ -231,14 +247,25 @@ def from_dict(cls, d: dict) -> "Excerpt":
...
+        # support conversion from string to integer when loading from csv
+        int_fields = [k for k, v in cls_field_types.items() if v == int]
+        for field_name in int_fields:
+            input_val = input_args.get(field_name)
+            if isinstance(input_val, str):
+                if input_val == "":
+                    del input_args[field_name]
+                else:
+                    input_args[field_name] = int(input_val)
+
```
laurejt, line 268: Can you please document this and other excerpt-behavioral modifications in the merge logs?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,354 @@
...
+It takes two or more input files of excerpt or labeled excerpt data in CSV format,
+merges the excerpts, and outputs a CSV of the merged excerpt data.  All excerpts
+are included in the output, whether they were merged with any other records or not.
+
+Merging logic is as follows:
+- Excerpts are merged on the combination of page id and excerpt id
+- When working with two sets of labeled excerpts, records are merged on the
+  combination of page id, excerpt id, and poem id
+    - If the same excerpt has different identifications, both
+      labeled excerpts will be included in the output
```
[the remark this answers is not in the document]
rlskoeser: By difference here I mean the same excerpt has different poem ids - I guess technically that is the "label"? I'll update to explicitly reference the field I mean and use the term label.

My hope is that this script will be the basis for later merging (if we need scripted merging and not just a utility method).

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,354 @@
...
+    - If the same excerpt has duplicate identifications, they will be merged
+      into a single excerpt that includes both identification methods
```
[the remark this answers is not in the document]
rlskoeser: if manual id and refmatcha agree, they are merged into one labeled excerpt and the identification methods field includes both. I'll take a stab at revising, but if you're not happy with my revision then please suggest changes.

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,354 @@
...
+    # before joining, drop redundant fields that will be the same
+    # on any excerpt with matching page & excerpt id
+    other_join = other_df.drop(
+        "detection_methods", "ppa_span_start", "ppa_span_end", "ppa_span_text"
+    )
+    merged = df.join(other_join, on=join_fields, how="left")
+
+    # if notes_right is present, then we have notes coming from both sides
+    # of the join; combine the notes into a single notes field
+    if "notes_right" in merged.columns:
```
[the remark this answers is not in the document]
rlskoeser: I can't imagine a case where notes order would matter. They'll be combined in the order that input files are specified; I'll add a line to the documentation.

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -0,0 +1,354 @@
...
+            .add(pl.lit("\n"))
+            .add(pl.col("notes_right").str.strip_chars())
+            .str.strip_chars("\n")
+        ).drop("notes_right")
+
+    if "identification_methods_right" in merged.columns:
+        # use list set union method to merge values, ignoring nulls
+        # - if left value is null, use right side
+        # - if right value is null, use left
+        # - if both are non-null, combine
```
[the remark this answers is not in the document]
rlskoeser: yes - this is the logic I came up with to generate the union of identification methods when poem_id is the same for two different id methods

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: yes, that's probably closer, I'll update and turn the comments into a doc string as you suggest

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: yes - I don't think we want to merge those yet

src/corppa/poetry_detection/core.py
rlskoeser: Could you clarify what you're asking for here ? 

I should have probably noted in the pull request overview, but it was a late addition when I figured out I needed it.  I'll update the docstring for the method.

src/corppa/poetry_detection/merge_excerpts.py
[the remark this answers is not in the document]
laurejt: Thanks for the clarification! I have a suspicion this will also become clearer once I start writing the pipeline documentation; the real issue for me has been trying to move between some of the more general forms of the comments and what specifically this script will do.

For instance, there is no plan to use this to generally merge labeled excerpts with labeled excerpts, right? In particular, I'm thinking of the case of merging passim-identified excerpts with the other ones we have.

src/corppa/poetry_detection/core.py
laurejt: I was just asking that there'd be a note / bullet point on this in the git log when this gets merged to develop. Then again, maybe all of these behaviors just need to be documented in the pipeline document? Although that might be the wrong level of detail for the document. Given our timeline, hopefully it won't matter. Just worried about not being able to keep track of all the various changes being made to excerpts and how they might impact other points of development in the pipeline.

laurejt (approved):
Looks good to me!

PR #175, opened 2025-02-28
Add method for correcting transformed excerpts

laurejt:


src/corppa/poetry_detection/core.py
```diff
@@ -6,6 +6,8 @@
 from dataclasses import asdict, dataclass, field, replace
 from typing import Any, Optional
 
+from Bio.Align import PairwiseAligner
```
rlskoeser, line 9: You need to add this package to our dependencies somewhere in pyproject.toml - I think that's the main reason tests are failing

src/corppa/poetry_detection/core.py
```diff
@@ -198,6 +200,33 @@ def strip_whitespace(self) -> "Excerpt":
...
+        """
+        aligner = PairwiseAligner(
+            mismatch_score=-0.5,
+            gap_score=-0.5,
+            query_left_gap_score=0,  # no penalty for gaps to the left of the excerpt
+            query_right_gap_score=0,  # no penlty for gaps to the right of the excerpt
+        )
+        results = aligner.align(page_text, self.ppa_span_text)
+        # Use first resulting alignment even if there are more than one
+        alignment = results[0]
```
rlskoeser, line 219: possible to get no results? what happens if we accidentally pass the wrong page text?

src/corppa/poetry_detection/core.py
```diff
@@ -198,6 +200,33 @@ def strip_whitespace(self) -> "Excerpt":
...
+        the detection process, this method attemps to correct the excerpt such
+        that the returned excerpt has the indicies and text correpsonding to
+        the original page text (rather than the transformed version). The
+        correspondence is determined using the Needleman-Wunsch algorithm.
+        """
+        aligner = PairwiseAligner(
+            mismatch_score=-0.5,
+            gap_score=-0.5,
+            query_left_gap_score=0,  # no penalty for gaps to the left of the excerpt
+            query_right_gap_score=0,  # no penlty for gaps to the right of the excerpt
```
rlskoeser, line 215: wonder if these should be defined at a higher level as a dict, rather than buried in this method (likely to change? will we want to easily reference them?)

src/corppa/poetry_detection/core.py
```diff
@@ -198,6 +200,33 @@ def strip_whitespace(self) -> "Excerpt":
...
+        )
+        results = aligner.align(page_text, self.ppa_span_text)
+        # Use first resulting alignment even if there are more than one
+        alignment = results[0]
+        ppa_aligned_seqs = alignment.aligned[0]
+        new_start = ppa_aligned_seqs[0][0]
+        new_end = ppa_aligned_seqs[-1][1]
```
rlskoeser, line 222: would be helpful to have some brief comments here about what the data structure is here

test/test_poetry_detection/test_core.py
```diff
@@ -359,6 +360,72 @@ def test_strip_whitespace(self):
         )
         assert excerpt.strip_whitespace() == expected_result
 
+    @patch("corppa.poetry_detection.core.PairwiseAligner")
+    def test_page_excerpt(self, mock_pairwise_aligner):
```
rlskoeser, line 364: It's a little hard to follow what the tests are doing. What's the value of mocking here instead of using the actual aligner?

test/test_poetry_detection/test_core.py
```diff
@@ -359,6 +360,72 @@ def test_strip_whitespace(self):
...
+            ppa_span_text="If there be any\nthis affembly, any dear friend of Caefar's, to him I say, that\nBrutus's love to Caefar was no less than his.",
+            detection_methods={"passim"},
+        )
+
+        expected_result = Excerpt(
+            page_id="CB0126086107.0218",
+            ppa_span_start=429,
+            ppa_span_end=551,
+            ppa_span_text="If there be any\nthis affembly, any dear friend of Cæfar's, to him I ſay, that\nBrutus's love to Cæfar was no leſs than his.",
+            detection_methods={"passim"},
```
rlskoeser, line 424: cool to see a real example and see how the alignment and text was adjusted

src/corppa/poetry_detection/core.py
```diff
@@ -198,6 +200,33 @@ def strip_whitespace(self) -> "Excerpt":
             ppa_span_text=self.ppa_span_text.strip(),
         )
 
+    def correct_page_excerpt(self, page_text: str) -> "Excerpt":
```
rlskoeser, line 203: Any thoughts on how we would call this outside of the Excerpt class context? e.g. if I want to run in batches on a polars dataframe?

No need to refactor now, but still worth thinking about

rlskoeser (commented):
Thanks for sharing, cool to see how this is working.

src/corppa/poetry_detection/core.py
laurejt: There will always be an alignment, it could be bad though.

src/corppa/poetry_detection/core.py
laurejt: I don't believe these should be modifiable. Why would we change them?

src/corppa/poetry_detection/core.py
laurejt: is it worth linking the BioPython API itself?

test/test_poetry_detection/test_core.py
laurejt: This entirely to get around the numerous data structures that BioPython uses. See their [documentation]
(https://biopython.org/docs/dev/Tutorial/chapter_pairwise.html#sec-pairwise-aligner)

src/corppa/poetry_detection/core.py
laurejt: Fixed!

src/corppa/poetry_detection/core.py
rlskoeser: Good to know they won't change, that's great. I think it's still better to make the settings we're using more visible by declaring them at a higher level (I was thinking just a dict that you pass into the init as kwargs), but I'm fine with this.

src/corppa/poetry_detection/core.py
rlskoeser: good to know how it's operating and what it's likely to do with bad input data; let's leave this as is until we know if it's a problem for our use cases

src/corppa/poetry_detection/core.py
rlskoeser: yes, sure!

test/test_poetry_detection/test_core.py
rlskoeser: 😵‍💫

rlskoeser: thanks for adding comments

rlskoeser (approved):
I'm fine with this being merged when you're ready. If/when we want to adapt the text realignment for use in a different case, we can figure out how to make it more reusable or possible to run on a polars dataframe in batches.

src/corppa/poetry_detection/core.py
laurejt: I've added warning to the method description to document this issue / potential fix.

src/corppa/poetry_detection/core.py
laurejt: That still would make them a modifiable input... which I'd rather not; that said I'll go make them class-level variables.

laurejt: I take that back, class-level variables could give us headaches elsewhere. Therefore I'm going to leave this for now, but I'll leave a comment documenting the potential future change.

src/corppa/poetry_detection/passim/README.md
```diff
@@ -70,33 +74,29 @@ excerpts and one that does not. This is due to the fact that `passim` output fil
...
+Example usage:
 ```
-python get_passim_excerpts.py passim_page_results.jsonl passim_excerpts.csv
+    get_passim_results.py --ppa-passim-corpus ppa_passim.jsonl --ref-corpus ref.jsonl \
+        --passim-dir passim_output --page-results passim_page_results.jsonl \
+        --span-results passim_spans.csv
+    get_passim_results.py --ppa-passim-corpus ppa_passim.jsonl --ref-corpus ref_a.jsonl \
+        --ref-corpus ref_b.jsonl --passim-dir passim_output \
+        --page-results passim_page_results.jsonl --span-results passim_spans.csv \
+        --ppa-text-corpus ppa.jsonl.gz
```
rlskoeser, line 101: Doesn't need to change [now], but these are really long parameters - seems like an unwieldy way to run things (lot of room for error)

laurejt: Sure, they're long. If you've got shorter forms that still make it explicitly clear how the various files are different, i'm open to changing the names.

PR #179, opened 2025-03-04
Excerpt improvements

rlskoeser:
This PR includes updates and improvements to Excerpt objects to support work on the merge script #156 and includes some changes pulled from the PR for that script #174 

changes include:
- new method to return a dictionary of field names and field base types 
  - supporting method to get actual type from field type annotation
  - ~add caching to fieldnames and field_types class methods~ (removed: causes problems with inheritance)
- consolidate `from_dict` method 
   - move input to set conversion to a method
   - rewrite as a single inherited method
   - add type details to exception message when methods values are passed in as an unsupported type
 - use a define delimiter character when converting sets to and from delimited string
 - add and test new `from_excerpt` method

### DIFF
--- a/src/corppa/poetry_detection/core.py
+++ b/src/corppa/poetry_detection/core.py
@@ -95,7 +95,11 @@ def field_real_type(field_type) -> type:
         return origin
     # if Optional or Union, return the first non-none type
     ftypes = get_args(field_type)
-    return [arg for arg in ftypes if arg != types.NoneType][0]
+    if ftypes:
+        return [arg for arg in ftypes if arg != types.NoneType][0]
+
+    # if we get here, this is an input we can't handle
+    raise ValueError(f"Cannot determine real type for '{field_type}'")
 
 
 @dataclass(kw_only=True, frozen=True)
--- a/src/corppa/poetry_detection/core.py
+++ b/src/corppa/poetry_detection/core.py
@@ -2,6 +2,7 @@
 Custom data type for poetry excerpts identified with the text of PPA pages.
 """
 
+import functools
 import types
 from copy import deepcopy
 from dataclasses import asdict, dataclass, field, fields, replace
@@ -184,12 +185,14 @@ class Excerpt:
         return csv_dict
 
     @classmethod
+    @functools.cache
     def fieldnames(cls) -> list[str]:
         """Return a list of names for the fields in this class,
         in order."""
         return [f.name for f in fields(cls)]
 
     @classmethod
+    @functools.cache
     def field_types(cls) -> dict[str, Any]:
         """Return a dictionary of field names and corresponding types
         for this class."""
--- a/src/corppa/poetry_detection/core.py
+++ b/src/corppa/poetry_detection/core.py
@@ -2,9 +2,10 @@
 Custom data type for poetry excerpts identified with the text of PPA pages.
 """
 
+import types
 from copy import deepcopy
 from dataclasses import asdict, dataclass, field, fields, replace
-from typing import Any, Optional
+from typing import Any, Optional, TypeAliasType, Union, get_args, get_origin
 
 # Table of supported detection methods and their corresponding prefixes
 DETECTION_METHODS = {
@@ -77,6 +78,24 @@ class Span:
         """
         overlap = self.overlap_length(other, ignore_label=ignore_label)
         return overlap / max(len(self), len(other))
+
+
+def field_real_type(field_type) -> type:
+    """Return the base type for a dataclass field type annotation.
+    For unions or optional values, returns the first non-None type; for type
+    aliases, returns the original type.
+    """
+    # if it's a regular type, return unchanged
+    if isinstance(field_type, type):
+        return field_type
+    # for a type alias, return the original type
+    # e.g. for annotation of set[str] return the set type
+    origin = get_origin(field_type)
+    if isinstance(origin, type):
+        return origin
+    # if Optional or Union, return the first non-none type
+    ftypes = get_args(field_type)
+    return [arg for arg in ftypes if arg != types.NoneType][0]
 
 
 @dataclass(kw_only=True, frozen=True)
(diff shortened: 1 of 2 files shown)


src/corppa/poetry_detection/core.py
```diff
@@ -79,6 +81,49 @@ def overlap_factor(self, other: "Span", ignore_label: bool = False) -> float:
...
+    :attr:`LabeledExcerpt.identification_methods`.
+    """
+    # match case syntax equivalent to isinstance(input_val, list)
+    match input_val:
+        case list():  # format used by to_json
+            return set(input_val)
+        case str():  # format used by to_csv
+            return set(input_val.split(MULTIVAL_DELIMITER))
+        case set():
+            return set(input_val)
```
laurejt, line 122: I don't understand this case; if the input is a set, why would you make a nested set?

src/corppa/poetry_detection/core.py
```diff
@@ -79,6 +81,49 @@ def overlap_factor(self, other: "Span", ignore_label: bool = False) -> float:
...
+    ftypes = get_args(field_type)
+    if ftypes:
+        return [arg for arg in ftypes if arg != types.NoneType][0]
+
+    # if we get here, this is an input we can't handle
+    raise TypeError(f"Cannot determine real type for '{field_type}'")
+
+
+#: character to use when converting sets to and from delimited string
+MULTIVAL_DELIMITER = ";"
```
laurejt, line 107: Since the delimiter is presumably being used in fields that might be read (i.e., CSVs); can this be a two-character delimiter?
```suggestion
MULTIVAL_DELIMITER = "; "
```

src/corppa/poetry_detection/core.py
```diff
@@ -79,6 +81,49 @@ def overlap_factor(self, other: "Span", ignore_label: bool = False) -> float:
         return overlap / max(len(self), len(other))
 
 
+def field_real_type(field_type) -> type:
+    """Return the base type for a dataclass field type annotation.
+    For unions or optional values, returns the first non-None type; for type
+    aliases, returns the original type.
+    """
```
laurejt, line 88: I'm not sure I understand this... Why do you need this behavior? When is there inheritance occurring here?

src/corppa/poetry_detection/core.py
```diff
@@ -166,29 +212,30 @@ def fieldnames(cls) -> list[str]:
...
+    def field_types(cls) -> dict[str, Any]:
+        """Return a dictionary of field names and corresponding types
+        for this class."""
+        return {f.name: field_real_type(f.type) for f in fields(cls)}
+
+    @classmethod
+    def from_dict(cls, d: dict) -> "Excerpt":
         """
         Constructs a new Excerpt from a dictionary of the form produced by `to_dict`
         or `to_json`.
```
laurejt, line 225:
```suggestion
        or `to_csv`.
```

src/corppa/poetry_detection/core.py
laurejt: Is this just another case where dataclasses and mypy don't like each other? We can probably largely drop the details for mypy since it won't be able to work well anyway with how the excerpt data types are constructed anyway

src/corppa/poetry_detection/core.py
```diff
@@ -79,6 +81,49 @@ def overlap_factor(self, other: "Span", ignore_label: bool = False) -> float:
...
+
+    # if we get here, this is an input we can't handle
+    raise TypeError(f"Cannot determine real type for '{field_type}'")
+
+
+#: character to use when converting sets to and from delimited string
+MULTIVAL_DELIMITER = ";"
+
+
+def input_to_set(input_val: list | str | set) -> set:
```
laurejt, line 110: Does it make since for this to be called in post-initialization?

src/corppa/poetry_detection/core.py
rlskoeser: Hmm, that may be my mistake, but the test was passing so I thought it worked as expected. If it's already a set, should we just return?

I wondered about your concerns about deep copy / set mutation - I think it could be solved here, but I don't think it impacts us right now so I didn't worry about that.

src/corppa/poetry_detection/core.py
rlskoeser: I wondered about that briefly -  your call

src/corppa/poetry_detection/core.py
rlskoeser: It's not inheritance, it's that type annotations / type hints != types.   The field object returns the type annotation, e.g. something like `Optional[str]` or `set[str]` - I want to know that the type is `str` or `set`. These are the methods that let us get that information.

I tried to make that clear in the docstring and comments - is it any clearer? Suggestions on how to improve? Would it help to add an example?

src/corppa/poetry_detection/core.py
laurejt: Honestly, I think it's a good idea. Assuming we do, input sets should definitely be passed through (rather than being nested).

I don't think it matters enough, but we could explicitly use frozensets rather than sets within initialization. This would prevent mutation of the set object itself. Would it be more pythonic type-wise to say inputs are strings or Iterables? Any iterable can be converted into a (frozen) set.

src/corppa/poetry_detection/core.py
laurejt: I think I just got confused, I was thinking "base" as the "base-level" class of some inheritance structure.

I'm also fine with us updating the type annotations so that dataclasses does the right thing. I think there's just mismatch between dataclasses and mypy when the data type being constructed gets too complex.

laurejt: I think an example would help.

rlskoeser: ah, I see - base is probably the wrong term, it does have that connotation!

src/corppa/poetry_detection/core.py
rlskoeser: I agree - iterable would be more pythonic (but then string is ambiguous because strings are iterables of characters)

laurejt: Fair enough.

laurejt (approved):

PR #180, opened 2025-03-05
Revise chadwyck-healey metadata to output id instead of filename

rlskoeser:
For use with the poem dataset we need chadwyck healey ids rather than filenames. My preference is to only include the id instead of the filename since we can (generally) determine filename from id. Is there any case that we care about where this is a problem?


laurejt (approved):
That seems fine with me, we may want to add some file utils logic if we decided to have the files live in any kind of nested structure, but the IDs makes more since to me here.

Issue #111, opened 2024-11-01
setup sphinx documentation


rlskoeser:
## tasks
- [x] basic setup for sphinx
- [x] clean up formatting for existing docstrings
- [ ] #186
- [ ] #187

laurejt:
As of 3/10/25:

- Added basic support for external linking, currently links to Python docs
- Fixed existing docs to get rid of all warnings during building of docs
- Updated typing & documentation for `path_utils.py` and `gvision_ocr.py` scripts
- Added filter script to docs, _but have not cleaned up typing and documentation_

mnaydan:
Decision at iteration planning to merge the version that we have as of 3/17

PR #182, opened 2025-03-06
Set up sphinx code documentation 

rlskoeser:
Work related to #111 

Compiling docs currently outputs a number of warnings with format errors in our docstrings, which we didn't know about as we wrote them because we weren't generating docs.


rlskoeser:
@laurejt your revisions here look good, is this ready to merge?

Issue #189, opened 2025-03-17
test merge script on passim excerpts + another set to see what it does


mnaydan:

rlskoeser:
completed in #196

PR #196, opened 2025-03-17
Merge labeled and unlabeled excerpts

rlskoeser:
Modifies merge script to handle more cases and simplify the code; addresses #189 

Revised to load all input files into a single dataframe, and then merge based on groupings of page id + excerpt id.  I was able to completely remove the old method that was doing the joins, and updated the existing tests to confirm the we get the expected results from the revised method.  

Here's the summary output from running it on all the adjucation data I have locally:
```console
 ❯ ./src/corppa/poetry_detection/merge_excerpts.py finish-line/adjudication/round1_excerpts.csv  \
  finish-line/adjudication/round1_excerpts_matched.csv finish-line/adjudication/round2_excerpts_merged.csv  \
  finish-line/adjudication/round2_excerpts.csv finish-line/adjudication/round2_excerpts_matched.csv \
  -o merge-all.csv
Loaded 4,264 excerpts from 5 files (1,090 labeled).
2,565 excerpts after merging; 876 labeled excerpts.
Total by detection method:
        adjudication: 2,565
        manual; refmatcha: 40
        manual: 173
        refmatcha: 663
```

I added the merge script to sphinx docs so that I could make sure the docstring was formatted correctly (it wasn't). I think we may want to make a separate section for the poetry detection code, but putting it at the top level for now.

question:
- do you think it is important now to setup parametrized unit tests so we can easily add test cases with expected number of resulting rows after merging? some of the existing tests could shift to this structure

### DIFF
--- a/src/corppa/poetry_detection/merge_excerpts.py
+++ b/src/corppa/poetry_detection/merge_excerpts.py
@@ -11,9 +11,9 @@ This means that in most cases, the output will likely be a mix of labeled
 and unlabeled excerpts.
 
 Merging logic is as follows:
-- Excerpts are merged on the combination of page id and excerpt id
-- When working with two sets of labeled excerpts, records are merged on the
-  combination of page id, excerpt id, and poem id
+- Excerpts are grouped on the combination of page id and excerpt id and then
+  merged if all reference fields match exactly, or where reference fields are
+  present in one excerpt and null in the other.
     - If the same excerpt has different labels (different `poem_id` values), both
       labeled excerpts will be included in the output
     - If the same excerpt has duplicate labels (i.e., the same `poem_id` from two
@@ -21,18 +21,9 @@ Merging logic is as follows:
       into a single labeled excerpt; the `identification_methods` in the
       resulting labeled excerpt will be the union of methods in the merged excerpts
 - When merging excerpts where both records have notes, the notes content
-  will be combined. (Combined notes order follows input file order.)
-
-After all input files are combined, the script checks for duplicate
-excerpt idenfications that can be consolidated. This currently only handles
-these simple cases:
-- All poem identification and reference fields match (poem_id, ref_span_start, ref_span_text, ref_span_end)
-- Poem identification matches and reference fields are null in one set
-    (e.g. manual identification, which does not include reference fields, and
-    refmatcha identification)
+  will be combined.
 
 Limitations:
-- Generally assumes excerpts do not require merging within a single input file
 - Merging based on poem_id does not compare or consolidate reference span indices
   and text fields; supporting multiple identification methods that output
   span information will require revision
@@ -82,16 +73,18 @@ def combine_duplicate_methods_notes(repeats_df: pl.DataFrame) -> pl.DataFrame:
 
 
 def merge_excerpts(df: pl.DataFrame) -> pl.DataFrame:
-    """Takes a polars Dataframe that includes labeled excerpts and merges:
-    - unlabeled excerpts with matching labeled excerpts (merge based on
-      page id and excerpt id)
-    - labeled excerpts with other labeled excerpts with the same label
-      (merge based on combination of page id, excerpt id, and poem id)
-
-    Returns the a dataframe with duplicate excerpts merged.
+    """Takes a polars Dataframe that includes labeled excerpts and merges.
     For now, merging is only done on the simple cases where reference
     fields match exactly, or where reference fields are present in one labeled
-    excerpt and unset in the other.
+    excerpt and unset in the other:
+    - unlabeled excerpts with matching labeled excerpts
+    - multiple labeled excerpts with the same label
+
+    When excerpts are merged, the detection_methods, identification_methods,
+    and notes fields are all combined to preserve all information.
+
+    Returns the a dataframe with duplicate excerpts merged.
+
     """
 
     # copy the df and add a row index for removal
@@ -134,34 +127,6 @@ def merge_excerpts(df: pl.DataFrame) -> pl.DataFrame:
                 ~pl.col("index").is_in(repeats.select(pl.col("index")))
             )
             # add the consolidated row with combined values to the merged df
-            merged_excerpts.extend(repeats[:1])
-
-    # merge labeled excerpts with matching labels
-    # group by page id, excerpt id, and poem id to find repeated identifications
-    for group, data in updated_df.group_by(["page_id", "excerpt_id", "poem_id"]):
-        # group is a tuple of values for page id, excerpt id, poem id
-        # data is a df of the grouped rows for this set
-
-        # sort so any empty values for optional fields are last,
-        # then fill values forward - i.e., treat nulls as duplicates
-        data = data.sort(
-            "ref_span_start", "ref_span_end", "ref_span_text", nulls_last=False
-        ).select(pl.all().backward_fill())
-
-        # identify repeats where reference values all agree
-        # (either same values or don't conflict because unset)
-        repeats = data.filter(
-            data.drop("identification_methods", "notes", "index").is_duplicated()
-        )
-
-        if not repeats.is_empty():
-            repeats = combine_duplicate_methods_notes(repeats)
-            # remove the repeats from the main dataframe
-            updated_df = updated_df.filter(
-                ~pl.col("index").is_in(repeats.select(pl.col("index")))
-            )
-            # add the consolidated row with combined values to the merged df
-            merged_excerpts.extend(repeats[:1])
             merged_excerpts.extend(repeats[:1])
 
     # combine and return
--- a/src/corppa/poetry_detection/merge_excerpts.py
+++ b/src/corppa/poetry_detection/merge_excerpts.py
@@ -1,37 +1,42 @@
 #!/usr/bin/env python
 """
-This script merges detected poems excerpts (:class:`~corppa.poetry_detection.core.Excerpt`)
-with identified poem excerpts (:class:`~corppa.poetry_detection.core.LabeledExcerpt`);
-it also handles merging duplicate poem identifications in simple cases.
+This script merges labeled and unlabeled poem excerpts, combining
+notes for any merged excerpts, and merging duplicate poem identifications 
+in simple cases.
 
-It takes two or more input files of excerpt or labeled excerpt data in CSV format,
-merges the excerpts, and outputs a CSV of the merged excerpt data.  All excerpts
-are included in the output, whether they were merged with any other records or not.
-This means that in most cases, the output will likely be a mix of labeled
-and unlabeled excerpts.
+It takes two or more input files of excerpt data (labeled or unlabeled) in CSV format,
+merges any excerpts that can be combined, and outputs a CSV with the updated excerpt data.  
+All excerpts in the input data files are preserved in the output, whether 
+they were merged with any other records or not. This means that in most cases, 
+the output will likely be a mix of labeled and unlabeled excerpts.
 
 Merging logic is as follows:
 
-- Excerpts are grouped on the combination of page id and excerpt id and then merged if all reference fields match exactly, or where reference fields are present in one excerpt and null in the other.
+- Excerpts are grouped on the combination of page id and excerpt id,
+  and then merged if all refence fields match exactly, or where 
+  reference fields are present in one excerpt and unset in the other.
     - If the same excerpt has different labels (different `poem_id` values), both
       labeled excerpts will be included in the output
     - If the same excerpt has duplicate labels (i.e., the same `poem_id` from two
       different identification methods), they will be merged
       into a single labeled excerpt; the `identification_methods` in the
       resulting labeled excerpt will be the union of methods in the merged excerpts
-- When merging excerpts where both records have notes, the notes content
+- When merging excerpts where both records have notes, notes content
   will be combined.
 
 Example usage:
 
-``./src/corppa/poetry_detection/merge_excerpts.py adjudication_excerpts.csv labeled_excerpts.csv -o merged_excerpts.csv``
+``./src/corppa/poetry_detection/merge_excerpts.py adjudication_excerpts.csv \
+    labeled_excerpts.csv -o merged_excerpts.csv``
 
 Limitations:
 
-- Merging based on poem_id does not compare or consolidate reference span indices
-  and text fields; supporting multiple identification methods that output
-  span information will require revision
+- Labeled excerpts with the same poem_id but different reference data
+  will not be merged; supporting multiple identification methods that output
+  span information will likely require more sophisticated merge logic
 - CSV input and output only (JSONL may be added in future)
+- Notes from an unlabeled excerpt are not currently merged to all corresponding
+  labeled excerpts (TODO: confirm with test case)
 
 """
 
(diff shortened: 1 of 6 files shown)


src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,152 +36,101 @@
...
-    # before joining, drop redundant fields that will be the same
-    # on any excerpt with matching page & excerpt id
-    other_join = other_df.drop(
-        "detection_methods", "ppa_span_start", "ppa_span_end", "ppa_span_text"
+    # get id methods as a list of lists, use itertools to unwrap
+    # the lists; consume the itertools generator and use set to uniquify,
+    # then convert back to list to put back in the polars dataframe
+    detect_methods = repeats_df["detection_methods"].drop_nulls().to_list()
+    combined_detect_methods = list(
+        set(list(itertools.chain.from_iterable(detect_methods)))
```
laurejt, line 59:
```suggestion
        set(itertools.chain.from_iterable(detect_methods))
```
`set` constructors can take iterables as input, so no need to transform it into a list first.

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,152 +36,101 @@
...
-        merged = merged.select(LABELED_EXCERPT_FIELDS).extend(
-            standardize_dataframe(right_df)
-        )
-
-    return merged
+    # id methods could be all unset even in a group
+    id_methods = repeats_df["identification_methods"].drop_nulls().to_list()
+    combined_id_methods = None
+    if id_methods:
+        combined_id_methods = list(set(list(itertools.chain.from_iterable(id_methods))))
```
laurejt, line 65:
```suggestion
        combined_id_methods = list(set(itertools.chain.from_iterable(id_methods)))
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,152 +36,101 @@
...
+    )
+    return repeats_df
 
 
-def merge_labeled_excerpts(df: pl.DataFrame) -> pl.DataFrame:
-    """Takes a polars Dataframe that includes labeled excerpts and attempts to
-    merges excerpts with same page id, excerpt id, and poem id. Returns the resulting
-    dataframe, with any duplicate excerpts merged.
+def merge_excerpts(df: pl.DataFrame) -> pl.DataFrame:
+    """Takes a polars Dataframe that includes labeled excerpts and merges.
```
laurejt, line 80: Random question: Can it merge undetected excerpts? Or, is it a requirement for one of the input dataframes to include labeled excerpts?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,152 +36,101 @@
...
     fields match exactly, or where reference fields are present in one labeled
-    excerpt and null in the other.
+    excerpt and unset in the other:
+    - unlabeled excerpts with matching labeled excerpts
+    - multiple labeled excerpts with the same label
+
+    When excerpts are merged, the detection_methods, identification_methods,
+    and notes fields are all combined to preserve all information.
+
+    Returns the a dataframe with duplicate excerpts merged.
```
laurejt, line 90:
```suggestion
    Returns a dataframe with duplicate excerpts merged.
```

src/corppa/poetry_detection/merge_excerpts.py
laurejt, line 80:
```suggestion
    """Takes a polars DataFrame that includes labeled excerpts and merges.
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,152 +36,101 @@
...
 
     # copy the df and add a row index for removal
     updated_df = df.with_row_index()
     # create a df with the same schema but no data to collect merged excerpts
     merged_excerpts = updated_df.clear()
 
-    # group by page id, excerpt id, and poem id to find repeated identificatins
-    for group, data in updated_df.group_by(["page_id", "excerpt_id", "poem_id"]):
+    # merge unlabeled excerpts with matching labeled excerpt
+    # OR excerpt with and without notes
```
laurejt, line 100: Does "excerpt" here refer to unlabeled excerpts?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,152 +36,101 @@
...
+    # group by page id and excerpt id only
+    for group, data in updated_df.group_by(["page_id", "excerpt_id"]):
         # group is a tuple of values for page id, excerpt id, poem id
         # data is a df of the grouped rows for this set
 
-        # sort so any empty values for optional fields are last,
-        # then fill values forward - i.e., treat nulls as duplicates
+        # sort so any empty values for optional fields are first,
```
laurejt, line 106: Just double checking, this is just optional fields relevant to identified excerpts, right? The notes field is also optional...

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,152 +36,101 @@
...
+            "ref_corpus",
+            "ref_span_start",
+            "ref_span_end",
+            "ref_span_text",
+            nulls_last=False,
+        ).select(pl.all().backward_fill())
+
+        # identify repeats where everything is the same but the row index,
+        # methods, and notes
+        # (other values must either be the same or don't conflict because they were unset)
```
laurejt, line 120: thanks for this clarifying parenthetical!

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,152 +36,101 @@
...
         repeats = data.filter(
-            data.drop("identification_methods", "index").is_duplicated()
+            data.drop(
+                "index", "detection_methods", "identification_methods", "notes"
+            ).is_duplicated()
         )
```
laurejt, line 125: What is the behavior if there are 3 excerpts, where 2 have conflicting fields but would "match" the third?

I'm not entirely sure what we'd expect the behavior to do: One idea is that the unset excerpt should merge with all possible excerpts in the group... but I'm not entirely sure.

The notes field isn't the highest priority, but it might be worth revisiting it's form; should it actually be a list/set of notes? Then again, this is likely overkill. It's not worth over-engineering a "nice to have" feature

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -236,27 +180,32 @@ def main():
         sys.exit(-1)
 
     total_excerpts = 0
```
laurejt, line 182: Is worth moving this code logic (from L238-L292) into its own method?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -236,27 +180,32 @@ def main():
...
-        # on the first loop, main excerpts df is unset, nothing to merge
-        if excerpts is None:
-            excerpts = merge_df
-        else:
-            # on every loop after the first, update excerpts by
-            # merging with the new input file
-            excerpts = combine_excerpts(excerpts, merge_df)
+    # combine input dataframes with a "diagonal" concat, which aligns
+    # columns and fills in nulls for missing columns in any of the dataframes
+    excerpts = pl.concat(input_dfs, how="diagonal")
```
laurejt, line 196: nice! I hadn't known about "diagonal" concats

src/corppa/poetry_detection/polars_utils.py
```diff
@@ -44,6 +44,8 @@ def fix_data_types(df):
             # only split if column is currently a string
             if df.schema[c] == pl.String:
                 df = df.with_columns(pl.col(c).str.split(MULTIVAL_DELIMITER))
+            elif df.schema[c] == pl.Null:
```
laurejt, line 47: Could you add a quick comment on what this case is doing? My understanding is that it's converting a null field for a list type into an empty list... Is that correct?

test/test_poetry_detection/test_merge_excerpts.py
```diff
@@ -76,25 +74,25 @@ def test_combine_excerpts_1ex_1label():
...
-    merged = combine_excerpts(df, other_df)
-    # expect two rows with  two different labels
+    df = pl.from_dicts(
+        [excerpt1.to_dict(), excerpt1_label1.to_dict(), excerpt1_label2.to_dict()]
+    )
+    merged = merge_excerpts(df)
+    # expect two rows with two different labels
     assert len(merged) == 2
-    # polars preserves order, so we can check that they match what was sent in
+    # original orderis NOT preserved due to grouping and sorting
```
laurejt, line 85:
```suggestion
    # original order is NOT preserved due to grouping and sorting
```

test/test_poetry_detection/test_polars_utils.py
```diff
@@ -106,6 +106,20 @@ def test_fix_datatypes():
...
+        {
+            "poem_id": ["a1", "a2", "a3"],
+            "ppa_span_start": ["1", "2", "3"],
+            "detection_methods": [None, None, None],
+        }
+    )
+    # if no values are set in a list field, type is pl.Null
+    fixed_df = fix_data_types(df)
+    # type should be set to list of string
+    assert fixed_df.schema["detection_methods"] == pl.List(pl.String)
```
laurejt, line 120: Should this test case be testing null identification methods? Its not valid for excerpt data to have null detection methods... or, does this have something to do with refmatcha's output form?

laurejt (approved):
Overall, I think it looks good. Most of my comments are updating documentation and documenting the behavior of unhandled cases.

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -1,196 +1,197 @@
...
+they were merged with any other records or not. This means that in most cases,
+the output will likely be a mix of labeled and unlabeled excerpts.
 
 Merging logic is as follows:
-- Excerpts are merged on the combination of page id and excerpt id
-- When working with two sets of labeled excerpts, records are merged on the
-  combination of page id, excerpt id, and poem id
+
+- Excerpts are grouped on the combination of page id and excerpt id,
+  and then merged if all refence fields match exactly, or where
```
laurejt, line 16:
```suggestion
  and then merged if all reference fields match exactly, or where
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,28 +47,31 @@
...
 
 
 def combine_duplicate_methods_notes(repeats_df: pl.DataFrame) -> pl.DataFrame:
-    """Takes a dataframe of repeated excerpts with duplicate information,
-    and combines detection_methods, identification_methods, and notes.
-    Returns the updated dataframe with the combined fields.
     """
-    # get id methods as a list of lists, use itertools to unwrap
+    Takes a dataframe of repeated excerpts with duplicate information,
+    and updateds all rows with combined set of unique
```
laurejt, line 59:
```suggestion
    and updates all rows with the combined set of unique
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -41,28 +47,31 @@
...
 from corppa.poetry_detection.polars_utils import load_excerpts_df, standardize_dataframe
 
 
 def combine_duplicate_methods_notes(repeats_df: pl.DataFrame) -> pl.DataFrame:
-    """Takes a dataframe of repeated excerpts with duplicate information,
-    and combines detection_methods, identification_methods, and notes.
-    Returns the updated dataframe with the combined fields.
     """
-    # get id methods as a list of lists, use itertools to unwrap
+    Takes a dataframe of repeated excerpts with duplicate information,
```
laurejt, line 58: Does this assume all excerpts (rows) in the dataframe correspond to the same (possibly labeled) excerpt? To me, repeated excerpts sounds more general like something containing duplicates, but I don't think that's not what is meant here.

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -185,6 +242,8 @@ def main():
     # load files and combine into a single excerpt dataframe
     for input_file in args.input_files:
         try:
+            # NOTE: very important to standardize input so that
+            # extraneous columns do not prevent duplicate excerpts from merging
```
laurejt, line 246: Thank you for adding this note!

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -193,17 +252,22 @@ def main():
 
     # combine input dataframes with a "diagonal" concat, which aligns
     # columns and fills in nulls for missing columns in any of the dataframes
-    excerpts = pl.concat(input_dfs, how="diagonal")
-    # get initial totals before merging
+    # NOTE: very important to standardize columns so that extraneous input
+    # columns do not prevent duplicate excerpts from merging
```
laurejt, line 256: It just occurred to me... does it make sense for at least some of this standardization to actually occur within `load_excerpts_df`?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -193,17 +252,22 @@ def main():
...
+    excerpts = standardize_dataframe(pl.concat(input_dfs, how="diagonal"))
+    # get initial totals before any uniquifying or merging
     total_excerpts = excerpts.height
+    # use unique to drop exact duplicates (passim results include exact dupes)
+    excerpts = excerpts.unique()
```
laurejt, line 261: Interesting... Do you have any examples, so I can check what might be going on? I hadn't expected this behavior, more anticipated the many cases where passim might identify multiple poems for the same PPA excerpt because of reference corpus duplication.

src/corppa/poetry_detection/polars_utils.py
```diff
@@ -41,9 +41,14 @@ def fix_data_types(df):
     for c, ctype in df_types.items():
         # For list (set) types, split strings on multival delimiter to convert to list
         if ctype is pl.List:
-            # only split if column is currently a string
+            # if excerpt content is loaded from csv, it will have an inferred
+            # type of string; split string content on our deliminter and
```
laurejt, line 45:
```suggestion
            # type of string; split string content on our delimiter and
```

laurejt (approved):

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: This is meant to be run on a set of excerpts that have already been determined to be duplicates. Is the word "repeated" tripping you up? Maybe rewording to call it duplicate excerpts would be clearer.

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: ah, whoops, did I add the comment in two places? I was debating about where to run the standardize and decided to do it once instead of in every file

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: I wondered about that too - we already have a check to see whether it's excerpt or labeled excerpt content, I think doing this properly would require a parameter to the standardize method to let it know which set of fields we need.

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: I'm wrong about this! Not sure why I thought this was the case, but I  just checked directly using polars `unique()` and both of the passim excerpt files have all rows unique.

src/corppa/poetry_detection/merge_excerpts.py
laurejt: In `load_excerpt_df` we already have logic that checks what columns are missing... so it seems like we could just select the expected fields and return that dataframe instead

src/corppa/poetry_detection/merge_excerpts.py
laurejt: yeah, it was just the term repeated. I just had been missing the assumption that all these rows will be fully merged into a singular row.

src/corppa/poetry_detection/merge_excerpts.py
laurejt: well, the duplicate comment did help point out that `load_excerpts_df` and `standardize_df` are doing different but in some ways overlapping things. Or rather, I'd have expected loading excerpts data to only keep excerpt-relevant columns, but standardization would be doing polars specific things... maybe?

Issue #152, opened 2025-02-10
create a notebook for data exploration


mnaydan:

rlskoeser:
Previously completed

PR #200, opened 2025-03-20
Preliminary review notebook

rlskoeser:
work towards #152 

in this PR:
- new top-level notebooks directory with a drafty version of a review notebook
- python dependencies for running and checking notebooks
- added a github actions workflow for an automated check that the notebook compiles, so we'll be alerted if we change anything in a way that prevents it from running

I know I need to add some high-level documentation about what this is and how to use it but wasn't sure where to put it or what it needed to include.  Main readme? Add a notebooks readme and link from main?

There are lots of todos — hopefully we can use to create, prioritize, and assign subtasks.

I'm using jupyter-lab to run the notebook locally from my dev environment and referencing a local clone of the gitlab data repo, which should be configurable - maybe easy way would be an environment variable?


rlskoeser:
🤦‍♀️ doh! notebook doesn't pass the github actions check because it doesn't have any data

@laurejt let me know your thoughts on this after you review; we could add some small fixture data for the check, maybe - the goal of the treon check is to make sure the code in the notebook runs

laurejt:
@rlskoeser I think we need to include some additional top-level documentation, because we need the minimal set-up to get everything to run and now there's a need to install gitlab data into a specific for now hard-coded location.

rlskoeser:
> @rlskoeser I think we need to include some additional top-level documentation, because we need the minimal set-up to get everything to run and now there's a need to install gitlab data into a specific (for now) hard-coded location.

Yes, I probably should have made my question in the PR description more clearly: I would like your input on what minimal documentation is needed and where you think it should go.  I know some docs are needed before we can merge.  

In case I haven't given enough details for you to to try out locally:
- clone a copy of the gitlab data repo
- install dev dependencies to get the jupyter dependencies: `pip install -e ".[dev]"`
- run jupyter-lab from the top-level of the directory and open the notebook

Go ahead and modify the path to the data directory for your checkout for now.

laurejt (commented):
I was able to get the notebook running locally via `notebook`. What about a modifiable config file for the notebook for anything that may vary by user?

laurejt:
@rlskoeser when it comes to documentation, I think we need to revisit what should be in the top-level README entirely and how various subdirectories should get their own more detailed READMEs, there's just too much to include it all in one place.

rlskoeser:
@laurejt updated to use the new config file (review #210 first), with sample data so treon can check the notebook.

I also added readme to the notebooks folder with notes on setup/install, how to start jupyter-lab, and notes on the treon / CI setup.

notebooks/poetry_excerpt_review.ipynb
laurejt, line 1: We should raise a `ValueError` if `data-dir` is not found

laurejt, line 1: The notebook currently crashes because `data_file.stem` only removes one of the two extensions for `excerpts.csv.gz`

laurejt: Also, as we talked about, we should add to the sample config that the poem dataset should be an absolute path

laurejt (changes requested):
Right now, the notebook currently crashes for me because it cannot find the excerpts file.

notebooks/poetry_excerpt_review.ipynb
rlskoeser: ah drat! I obviously didn't test that case. Is there a good pathlib solution for this?

laurejt: perhaps `suffixes` this will get a list of the extensions for a Path

laurejt (approved):
I can run the notebook locally, so it's good enough to merge.

Issue #204, opened 2025-03-21
create a standard way within corppa to combine poetry metadata and text from all reference corpora


mnaydan:
- n/a tiny corpus
- Internet Poems
- Chadwyck-Healey

primary goal / must have:
- [x] add tiny corpus
- [x] method to generate merged poem metadata for use with excerpts

revise configurations / setup for reuse:
- [x] tar.gz configs should be text_dir and allow EITHER tarball or dir path
- [x] internet poems uses tarball file list to generate metadata (or dir path)
- [x] other poems config `metadata_path` is url or local file
- [x] update shared config file
- [x] simplify polars schema handling (disable inference)
- [x] raise notimplemented if tar file not handled

review/testing notes:

- [x] able to run locally
- [x] able to run on prosody staging vm

follow up work (later):
- #221

rlskoeser:
resolved by #258

Issue #201, opened 2025-03-20
revise refmatcha reference text/metadata compilation to use local, filtered CH content


rlskoeser:

rlskoeser:
minimal updates for this change merged into develop in #208

PR #208, opened 2025-03-21
Update refmatcha to use local, filtered version of chadwyck-healey metadata & text

### LINKED ISSUE #190: standardize corpus IDs
- ~ create a define or enum in poetry detection core so we can use the definitions elsewhere instead of strings~
- ~ document known ids for use in the data dictionary~
- [x] update refmatcha to use the internet poems identifier used with passim
- [x] recompile and rerun refmatcha 
- [x] put a copy of the results on google drive
- [x] rerun merge script and update excerpts in data repo

rlskoeser:
This is the minimal set of changes I needed to get refmatcha working with the new CH metadata.  Recompiling the new metadata means refmatcha outputs will now have the updated Internet Poems corpus id, so this resolves #190 as well as #201

I added a `--recompile` option because it was helpful for testing.

I'm getting different and slightly fewer matches than before - spent a little time trying to figure out why and didn't come up with anything conclusive. The CH ids for the matches I've lost seem to be present in our filtered set, so I'm going to move forward with this.

I'll clean up & document the reference corpus stuff as part of #204

### DIFF
--- a/src/corppa/poetry_detection/refmatcha.py
+++ b/src/corppa/poetry_detection/refmatcha.py
@@ -171,7 +171,7 @@ def compile_metadata(data_dir, output_file):
             .with_columns(source=pl.lit(SOURCE_ID["Chadwyck-Healey"]))
             .with_columns(
                 pl.concat_str(
-                    [pl.col("author_fname"), pl.col("author_lname")],
+                    [pl.col("author_firstname"), pl.col("author_lastname")],
                     separator=" ",
                 ).alias("author")
             )
@@ -249,12 +249,8 @@ def _text_for_search(expr):
         expr.str.replace_all(r"(\w)[-'](\w)", "$1$2")
         # metrical notation that splits words (e.g. sudden | -ly or visit | -or)
         .str.replace_all(r"(\w) \| -(\w)", "$1$2")
-        # replace other puncutation with spaces
+        # replace other punctuation with spaces
         .str.replace_all("[[:punct:]]", " ")
-        # remove indent entity in CH (probably more of these...)
-        .str.replace_all(
-            "&indent;", " "
-        )  # TODO: remove when we switch to new version of CH texts
         .str.replace_all(
             LINE_SEPARATOR, "\n"
         )  # replace unicode line separator with newline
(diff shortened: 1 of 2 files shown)


src/corppa/poetry_detection/refmatcha.py
```diff
@@ -157,6 +157,8 @@ def compile_metadata(data_dir, output_file):
     # open a parquet writer for outputting content in batches
     pqwriter = pq.ParquetWriter(output_file, schema)
 
+    # TODO: prioritize internet-poems matches over CH
+
```
laurejt, line 161: It seems fine to leave this as a TODO, I assume the current behavior is just whatever is encountered first?

src/corppa/poetry_detection/refmatcha.py
```diff
@@ -341,10 +339,10 @@ def multiple_matches(filtered_ref_df):
...
     # if author/title duplication check failed, check for author matches
-    # poetry foundation includes shakespeare drama excerpts with alternate names
+    # poetry foundation includes Shakespeare drama excerpts with alternate names
     authordupe_df = df.filter(df.select(["_author"]).is_duplicated())
     if not authordupe_df.is_empty():
-        # shakespeare shows up oddly in poetry foundation;
+        # Shakespeare shows up oddly in poetry foundation;
         # if author matches, assume the other source has the correct title
         non_poetryfoundtn = authordupe_df.filter(
             pl.col("source") != SOURCE_ID["Poetry Foundation"]
```
laurejt, line 348: Do we need this if we're not using Poetry Foundation?

laurejt (approved):

src/corppa/poetry_detection/refmatcha.py
rlskoeser: yes, right now the order things are added to the compiled reference data is the order they are chosen

src/corppa/poetry_detection/refmatcha.py
rlskoeser: I'll keep in mind when I work on the reference corpus metadata, I think you're right but it can be handled later

PR #209, opened 2025-03-21
Feature/utils docs

laurejt:
ref: #185 

Finished cleaning up and adding sphinx docs for all utility scripts


rlskoeser (approved):
Thanks, looks good

PR #210, opened 2025-03-21
Basic setup for local configuration

rlskoeser:
configuration setup for the notebook (and elsewhere) - related to #152 

Now updated with corrected paths and test paths now that I've tested it with the notebook.

In this pr:

- new `corppa.config` file with utility method `get_config`
- unit tests for loading config options and error handling  when config file is not present
- sample config file with one section and one path for now
- gitignore to keep the local config from being checked in
- added a note about config file to developer setup instructions 

I went with the python built in `configparser` since the toml support is read-only and we might eventually want more than that.

Note: I switched unit tests to use an editable pip install (`-e`) so that the configpath error test case matches local development; otherwise the path is relative to a package install instead of current working directory.

### DIFF
--- a/src/corppa/config.py
+++ b/src/corppa/config.py
@@ -5,23 +5,22 @@ Load local configuration options
 import configparser
 import pathlib
 
-#: src dir relative to this file (assumeing dev environment for now)
-CORPPA_SRC_DIR = pathlib.Path(__file__).parent.absolute()
+#: src dir relative to this file (assuming dev environment for now)
+CORPPA_SRC_DIR = pathlib.Path(__file__).parent.parent.absolute()
 
-#: expected patch for local config file (non-versioned)
+#: expected path for local config file (non-versioned)
 CORPPA_CONFIG_PATH = CORPPA_SRC_DIR.parent / "corppa.cfg"
-#: expected patch for example config file
+#: expected path for example config file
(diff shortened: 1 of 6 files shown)


src/corppa/config.py
```diff
@@ -0,0 +1,27 @@
...
+#: expected path for example config file
+SAMPLE_CONFIG_PATH = CORPPA_SRC_DIR.parent / "sample.cfg"
+
+
+def get_config():
+    # if the config file is not in place
+    if not CORPPA_CONFIG_PATH.exists():
+        not_found_msg = f""""Config file not found.
+Copy {SAMPLE_CONFIG_PATH} to {CORPPA_CONFIG_PATH} and configure for your environment.
+"""
```
laurejt, line 22: I I think there's an extra quote in here... but could you use in line quotes and make the newline explicit?

test/test_config.py
```diff
@@ -0,0 +1,30 @@
...
+import pytest
+
+from corppa import config
+
+
+def test_get_config_not_found(tmp_path):
+    test_config = tmp_path / "test.cfg"
+    # error should include directions about how to fix the problem
+    expected_error_msg = f"""Config file not found.
+Copy .*{config.SAMPLE_CONFIG_PATH.name} to .*{test_config.name} and configure for your environment."""
```
laurejt, line 12: Could you use in line quotes and make the newline explicit?

laurejt (approved):

PR #212, opened 2025-03-24
Switch config file to yaml so we can have nested configs

rlskoeser:
in this pr:
- changed config file and loading to yaml
  - updated sample and test config file format
  - renamed files and updated documentation to use the new names
- add exception handling and unit test for parse error


src/corppa/config.py
```diff
@@ -2,16 +2,22 @@
...
 """
 
-import configparser
 import pathlib
 
+import yaml
+
+try:
+    from yaml import CLoader as Loader
+except ImportError:
```
laurejt, line 11: Would it be reasonable to move the "pragma: no cover" to include the except clause too?

```suggestion
except ImportError:  # pragma: no cover
```

laurejt (changes requested):
The notebook crashes with the following error:
> `SystemExit: Error parsing config file: did not find expected <document start>` in <config file> at line 8, column 1.

The `corppa_config.yml` file exists, but is there any syntax that needs to be changed for the path?

src/corppa/config.py
rlskoeser: ah yep, that makes sense - thanks

rlskoeser:
> The notebook crashes with the following error:
> 
> > `SystemExit: Error parsing config file: did not find expected <document start>` in  at line 8, column 1.
> 
> The `corppa_config.yml` file exists, but is there any syntax that needs to be changed for the path?

Ah, sorry, I failed to update the sample config file with the new syntax. I think I missed it because I wanted to make sure I pulled in the note we added about the absolute path, but that was on the review notebook branch...  Should have both the new syntax and the note now.

laurejt (approved):
I can successfully run the notebook, this seems good enough for an initial merge.

PR #214, opened 2025-03-24
Consolidate reference corpora data and generate combined poem metadata file

rlskoeser:
ref: #204 

This consolidates scattered logic for working with and compiling reference corpus metadata and text, using the existing utility methods from the `build_text_corpus` script and adapting (but not yet replacing) logic from `refmatcha`.

I also created a preliminary `compile_dataset` script - for now it only handles the poem metadata.

I haven't written unit tests yet - wanted a check on the direction.


sample_config.yml
```diff
@@ -6,3 +6,14 @@ poem_dataset:
...
+# paths to data and metadata for
+reference_corpora:
+  internet_poems:
+    data_path: path/to/internet-poems/
+  chadwyck-healey:
+    data_path: path/to/chadwyck-healey/
+    metadata_path: path/to/chadwyck-healey_metadata.csv
+  other_poems:
+    # url for google sheets as CSV
+    metadata_url:
```
laurejt, line 19: I'm not opposed to using google sheets, but then it seems like we need to do this for the adjudicated excerpts as well. In that case, how do we keep track of what refmatcha is using vs. whatever might exist on google drive?

sample_config.yml
```diff
@@ -6,3 +6,14 @@ poem_dataset:
...
   data_dir: "/path/to/ppa-found-poems/data/"
+
+# paths to data and metadata for
+reference_corpora:
+  internet_poems:
+    data_path: path/to/internet-poems/
+  chadwyck-healey:
+    data_path: path/to/chadwyck-healey/
+    metadata_path: path/to/chadwyck-healey_metadata.csv
+  other_poems:
```
laurejt, line 17: Does it make more sense to have one singular directory, like with the `data_dir` (where we assume some organization?)

sample_config.yml
laurejt, line 1: How are you thinking of organizing PPA data within this?

sample_config.yml
rlskoeser: Agreed about adjudicated excerpts. Should `compile_dataset` script have a step to refresh local copies where it pulls down adjudicated excerpts? Should it update a local working copy or update the shadow dataset version? (both? Or do we even need a local copy if this is meant to be run with something that has access to the shadow dataset?)

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+class BaseReferenceCorpus:
+    """
+    Base class for reference poetry corpora, with corpus identifier and
+    methods to access metadata and text content.
+    """
+
+    corpus_id: str
+    corpus_name: str
```
laurejt, line 17: Is this just for reporting? Or do you see this as an additional field in our outputs?

sample_config.yml
rlskoeser: I hadn't thought that far yet. Can we set it up parallel to the shadow dataset structure?

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+        # load data path from config file and check that path exists
+        config_opts = get_config()
+        # TODO: handle missing config opts
+        return config_opts["reference_corpora"][self.corpus_id]
+
+    def get_metadata(self) -> Generator[dict[str, str]]:
+        """Minimal common poetry metadata for use across reference corpora.
+        Should yield a dictionary with poem_id, author, and title for each poem
+        in this corpus."""
+        raise NotImplementedError
```
laurejt, line 39: Is there a particular use case that needs a dictionary, as opposed to a `polars` DataFrame? Also, is it worth having a generator vs. just returning a whole DataFrame?

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+    Base class for reference poetry corpora, with corpus identifier and
+    methods to access metadata and text content.
+    """
+
+    corpus_id: str
+    corpus_name: str
+    data_path: pathlib.Path
+    metadata_path: pathlib.Path
+
+    def get_config_opts(self):
```
laurejt, line 21: missing return type

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+        self.data_path = pathlib.Path(config_opts["data_path"])
+        if not (self.data_path.exists() and self.data_path.is_dir()):
+            raise ValueError(
+                "Internet Poems Reference Corpus is not configured correctly"
+            )
+
+    def get_metadata(self) -> Generator[dict[str, str]]:
+        for text_file in self.data_path.glob("*.txt"):
+            # filename format:
+            #   Firstname-Lastname_Poem-Title.txt
```
laurejt, line 83: I'm willing to ignore my misgivings on assuming the metadata can be entirely derived from the names, but that's because I'm going to assume (a) this is fixed and that (b) we'll never correct any of its metadata.

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+        self.metadata_url = config_opts["metadata_url"]
+        # TODO: handle key error
+
+    def get_metadata(self):
+        # polars can load csv from a url;
+        meta_df = pl.read_csv(self.metadata_url)
+        # field are already named appropriately
+        yield from meta_df.iter_rows(named=True)
+
+    # this is a metadata-only corpus, so leave get_text as not implemented
```
laurejt, line 146: This is worth adding to the top-level description for the class

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def all_corpora():
```
laurejt: Missing type signature

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+    return [InternetPoems(), ChadwyckHealey(), OtherPoems()]
```
laurejt, line 152: Is the goal here just to return a list of reference corpora?

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+            .rename({"title_main": "title", "id": "poem_id"})
+            .with_columns(
+                pl.concat_str(
+                    [pl.col("author_firstname"), pl.col("author_lastname")],
+                    separator=" ",
+                ).alias("author")
+            )
+            .select(["poem_id", "author", "title"])
+        )
+        yield from df.iter_rows(named=True)
```
laurejt, line 125: Note there might be some typing issues here, since `polars` typing signature is `Iterator[tuple[Any, ...]] | Iterator[dict[str, Any]]`

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+            "ref_corpus": pl.String,
+        },
+    )
+
+    # for each corpus, load poem metadata into a polars dataframe,
+    # rename id to poem_id, and add a column with the corpus id
+    for ref_corpus in all_corpora():
+        corpus_meta = pl.from_dicts(ref_corpus.get_metadata()).with_columns(
+            ref_corpus=pl.lit(ref_corpus.corpus_id)
+        )
```
laurejt, line 180: Wouldn't it be easier just to get a DataFrame here?

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+    if not (output_data_dir.exists() and output_data_dir.is_dir()):
+        raise ValueError("Poem dataset path is not configured correctly")
+
+    output_file = output_data_dir / "poem_meta.csv"
+
+    # check & report if the file already exists
+    output_verb = "Creating"
+    if output_file.exists():
+        output_verb = "Replacing"
+    print(f"{output_verb} {output_file}")
```
laurejt, line 200: Is it worth saving / renaming the older version?

src/corppa/poetry_detection/ref_corpora.py
```diff
@@ -0,0 +1,184 @@
...
+        ref_corpus.corpus_id: ref_corpus.corpus_name for ref_corpus in all_corpora()
+    }
+
+    total_by_corpus = df["ref_corpus"].value_counts()
+    totals = []
+    for value, count in total_by_corpus.iter_rows():
+        # row is a tuple of value, count;  convert reference corpus id to name
+        totals.append(f"{ref_corpus_names[value]}: {count:,}")
+
+    print(f"{df.height:,} poem metadata entries ({'; '.join(totals)})")
```
laurejt, line 213: This is just some reporting, right?

sample_config.yml
laurejt: I assume so, it's actually contained within the shadow data directory; so if we're assuming its form than we can assume similar path constructions for the PPA data

sample_config.yml
laurejt: Just writing down what we discussed: we should have an option to override the assumed config path by providing a path during initialization.

src/corppa/poetry_detection/compile_dataset.py
```diff
@@ -0,0 +1,23 @@
+# compile the found poem dataset
+
+from corppa.config import get_config
+from corppa.poetry_detection.ref_corpora import save_poem_metadata
+
+
+def main():
+    config_opts = get_config()
+    required_sections = ["poem_dataset", "reference_corpora"]
```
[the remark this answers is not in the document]
rlskoeser: Yes, sorry about that, let's rename it.  Is `excerpts_dataset` clear enough? Or `compiled_dataset` ?

src/corppa/poetry_detection/ref_corpora.py
rlskoeser: reporting only for now

src/corppa/poetry_detection/ref_corpora.py
rlskoeser: Yes, I wanted an easy way to get the full set - and for refmatcha I want an easy way to get all the full-text corpora.

I know it can be done with inheritance but was going for something simple. LMK if you have thoughts.

src/corppa/poetry_detection/ref_corpora.py
rlskoeser: yes - for the compile dataset script. maybe it belongs there instead of here?

src/corppa/poetry_detection/compile_dataset.py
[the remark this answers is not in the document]
laurejt: How about compiled dataset, just because we technically have intermediate excerpt datasets.

src/corppa/poetry_detection/ref_corpora.py
laurejt: *shrug* I think it's fine, it might be something nice to have elsewhere.

src/corppa/poetry_detection/ref_corpora.py
laurejt: It's probably fine? My inclination is to have a meta variable with a list of corpus IDs, and a `get_corpora` method that builds the list based on an input list of corpus ids (`list[str]`).

src/corppa/poetry_detection/compile_dataset.py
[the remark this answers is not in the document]
rlskoeser: I'll switch to compiled dataset for the name but want to wait and do it as part of the compile dataset script. Just added a note to #216

src/corppa/poetry_detection/ref_corpora.py
rlskoeser: I say no, since this is in the data repo which uses git for versioning.  I'll have the compile dataset output a reminder to commit & push changes

laurejt (commented):
Posting my comments now, but I don't want to approve this until I can get it running this running locally.

sample_config.yml
```diff
@@ -7,13 +7,21 @@ poem_dataset:
...
-  other_poems:
+  base_dir: /tigerdata/cdh/prosody/excerpt-data-ingredients/ref-corpora/
+  # Uses default paths relative to reference_corpora.base_dir
+  # Uncomment to override if needed; if path is not absolute,
+  # will be assumed relative to reference_corpora base_dir
+  # internet_poems:
+    # data_path: internet_poems
+  # chadwyck-healey:
+    # data_path: chadwyck-healey/chadwyck-healey_texts
+    # metadata_path: chadwyck-healey/chadwyck-healey.csv
```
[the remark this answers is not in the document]
rlskoeser: ah, thanks

src/corppa/poetry_detection/compile_dataset.py
```diff
@@ -0,0 +1,23 @@
...
+    for section in required_sections:
+        if section not in config_opts:
+            print(f"Required configuration for '{section}' not found in config file")
```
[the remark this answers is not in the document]
rlskoeser: yes, will handle in the compile-dataset branch

src/corppa/poetry_detection/ref_corpora.py
rlskoeser: agree that would be nicer; I also wondered about a full_text flag on the reference corpus objects so we can filter; seems out of scope for now

Issue #194, opened 2025-03-17
add Github action to publish sphinx docs


mnaydan:

rlskoeser:
@laurejt here's the sphinx docs workflow for the PPA django codebase: https://github.com/Princeton-CDH/ppa-django/blob/main/.github/workflows/sphinx_docs.yml

We have a similar workflow in the S&co codebase https://github.com/Princeton-CDH/geniza/blob/main/.github/workflows/sphinx_docs.yml

That one includes a simple code coverage check.

In case another example is useful, in undate we have a check to make sure docs build without warnings https://github.com/dh-tech/undate-python/blob/main/.github/workflows/check.yml#L29-L32

PR #219, opened 2025-03-25
Add initial sphinx docs workflow

laurejt:
ref: #194, #187

### DIFF
--- a/.github/workflows/sphinx_docs.yml
+++ b/.github/workflows/sphinx_docs.yml
@@ -46,7 +46,7 @@ jobs:
       # For pull requests, exit with error if documentation coverage is incomplete
       - name: Report on documentation coverage
         if: ${{ github.event_name == 'pull_request' }}
-        run: cd docs && if [[ $((`grep corpppa build/coverage/python.txt | grep -v 100 | wc -l`)) -eq 0 ]] ; then echo "Documentation coverage complete"; else cat build/coverage/python.txt && exit 1; fi
+        run: cd docs && if [[ $((`grep corpppa build/coverage/python.txt | grep -v 100 | wc -l`)) -eq 0 ]] ; then cat build/coverage/python.txt && echo "" && echo "Documentation coverage complete"; else cat build/coverage/python.txt && exit 1; fi
 
       # When building on push to main, publish the built docs
       - name: Deploy built docs to github pages
--- a/.github/workflows/sphinx_docs.yml
+++ b/.github/workflows/sphinx_docs.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
       - develop
-    pull_request:
+  pull_request:
 
 env:
   # Python version used to build sphinx docs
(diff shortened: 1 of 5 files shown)


.github/workflows/sphinx_docs.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Install dependencies
+        run: pip install -e ".[dev]"
+
+      - name: Build Sphinx docs
+        run: cd docs && make -b coverage html
+
+      # For pull requests, exit with error if documentation coverage is incomplete
+      - name: Report on documentation coverage
+        if: ${{ github.event_name == 'pull_request' }}
+        run: if [[ $((`grep ppa _build/coverage/python.txt | grep -v 100 | wc -l`)) -eq 0 ]] ; then echo "Documentation coverage complete"; else cat sphinx-docs/_build/coverage/python.txt && exit 1; fi
```
rlskoeser, line 49: Should the `ppa` here be changed to `corppa` ? The paths might need some adjustment also. Please try running this command locally and see what needs to be changed.

rlskoeser: Might be something like this, but I'm guessing

```suggestion
        run: if [[ $((`grep corpppa _build/coverage/python.txt | grep -v 100 | wc -l`)) -eq 0 ]] ; then echo "Documentation coverage complete"; else cat docs/_build/coverage/python.txt && exit 1; fi
```

.github/workflows/sphinx_docs.yml
```diff
@@ -0,0 +1,57 @@
...
+          path: ~/.cache/pip
+          key: pip-${{ hashFiles('pyproject.toml') }}
+          restore-keys: |
+            pip-${{ hashFiles('pyproject.toml') }}
+            pip-
+
+      # TODO: Is installing the dev environment the right call?
+      # Install dependencies
+      - name: Install dependencies
+        run: pip install -e ".[dev]"
```
rlskoeser, line 41: It would be better if we didn't have to install all the dev dependencies; I'm not sure if package + doc optional dependencies is enough, if we have documentation for python code that imports optional dependencies.  Maybe worth trying to see what fails?

```suggestion
      # Install package + dependencies for building docs
      - name: Install dependencies
        run: pip install -e ".[doc]"
```

rlskoeser (commented):
This looks like a good start! I think some paths may need adjustment.

I'm not sure why the workflow didn't run yet... We might want to temporarily change it to run on more cases; maybe try enabling it to run on push to develop?

.github/workflows/sphinx_docs.yml
laurejt: Good point, I missed this reference. Thanks!

laurejt:
@rlskoeser one thing I'm less sure of is if we stop the check for failing for methods we want to ignore (i.e., cases where we only want to include top-level documentation)

rlskoeser (approved):
Great! Let's merge 🚀

PR #228, opened 2025-04-17
 Consolidate reference corpora data and generate combined poem metadata file (redux)

rlskoeser:
ref: #204 

Consolidates scattered logic for working with and compiling reference corpus metadata and text, using the existing utility methods from the `build_text_corpus` script and adapting (but not yet replacing) logic from `refmatcha`.

Also includes a preliminary `compile_dataset` script - in this branch it only handles the poem metadata.


rlskoeser:
This is working for me locally, I'm going to merge and we can address any problems later on as needed.

PR #232, opened 2025-04-28
Add EoP utilities doc

laurejt:
**Associated Issue(s):** #233 

### Changes in this PR

- Added new doc for Ends of Prosody utilities documentation
- Added this doc to the docs index so the page would compile via `make html`

### Questions
Below are the questions I have for this draft PR.

- Does the new document contain all the documentation we need?
- This should be a standalone document as opposed to part of the sphinx docs for the package, right? If that's the case, what's the proper way to go about building this standalone doc?

### DIFF
--- a/docs/source/eop-docs.rst
+++ b/docs/source/eop-docs.rst
@@ -1,0 +1,15 @@
+Ends of Prosody: corppa utilities
+#################################
+
+Filter Utility
+--------------
+.. automodule:: corppa.utils.filter
+   :noindex:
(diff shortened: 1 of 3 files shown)


rlskoeser (approved):
This is pretty much what I imagined.  I think it's fine.

I don't think we need to separate this out from the other sphinx docs, we just link them directly to this page with the most relevant documentation.   If you're concerned about anyone wandering over to the other docs, we could add a brief disclaimer to the top of the code docs, but I don't think it's necessary.

Issue #191, opened 2025-03-17
tidy branches and tag release


mnaydan:
- [x] enable zenodo deposit for this repository with cdh zenodo account
- [x] add license file and document license in readme
- [x] clean up any working github branches that we don't want to keep or merge in
   - [x] document experimental/historical branches briefly somewhere (readme maybe?)
- [x] create a release based on the tag to mint a DOI with Zenodo
- [x] add DOI badge to the main readme

PR #235, opened 2025-04-29
Add copywrite config file and add header to all files

rlskoeser:
**Associated Issue:** #191

### Changes in this PR

- I used [copywrite](https://github.com/hashicorp/copywrite) to add license & copyright to headers of all files
- Add copywrite config file for the project

### Notes

- I tried the pre-commit hook but couldn't get it to run; seems like it's been broken for a while https://github.com/hashicorp/copywrite/issues/78
- I added notes to the developer documentation about how I ran it - not sure it's worth installing/testing since the pre-commit hook apparently does not work and the tool does not update files that already have a copyright notice in the header.

### Evaluation

- [x] Review copyright statement
- [x] Review excluded files in copywrite config
- [x] Review developer notes about copywrite and confirm it is documented clearly enough


.copywrite.hcl
```diff
@@ -0,0 +1,16 @@
...
+project {
+  license        = "Apache-2.0"
+ # copyright_year = 2025  # NOTE: this doesn't seem to be used anywhere
+  copyright_holder = "2024,2025 Center for Digital Humanities, Princeton University"
+
+  # (OPTIONAL) A list of globs that should not have copyright/license headers.
+  # Supports doublestar glob patterns for more flexibility in defining which
+  # files or folders should be ignored
+  header_ignore = [
+    "docs/build/**"
```
laurejt, line 12: Just to verify: we want copyright statements on the other files within `docs`

laurejt (commented):
Should there be copyright notices on the Jupyter Notebooks?

DEVELOPERNOTES.md
```diff
@@ -43,3 +43,20 @@ This will create a file under `_build/coverage/python.txt` listing any Python cl
...
+A copyright notice has been applied to the header of all files with 
+[copywrite](https://github.com/hashicorp/copywrite) with the project configuration in `.copywrite.hcl`.
+
+To install with brew:
+```sh
+brew tap hashicorp/tap
+brew install hashicorp/tap/copywrite
+```
+
+Once installed, you can apply the copyright notice to all files without this notice by running `copywrite headers` (note that id does not update existing files even if the configuration has changed).
```
laurejt, line 58:
```suggestion
Once installed, you can apply the copyright notice to all files without this notice by running `copywrite headers` (note that it does not update existing files even if the configuration has changed).
```

.copywrite.hcl
rlskoeser: I think for any other supported files it would be reasonable? I just looked and I don't think it's added these anywhere except to python files.

PR #256, opened 2026-02-25
Merge exact spans and consolidate different poem ids

### LINKED ISSUE #255: make exact excerpt merging less conservative
Previous logic only merged if excerpt id matches (which includes id method) and if poem id matched; that leaves too many duplicate excerpts. Revise to merge anything with the exact same span in the PPA text, regardless of id method; consolidate different poem ids (in most cases they are likely near or full duplicate ids anyway).

rlskoeser:
**Associated Issue(s):** resolves #255

### Changes in this PR

- Revise existing merge logic so all excerpts with matching ppa spans are consolidated
- Rewrites merge logic to use polars rather than a loop (much faster!)

### Notes

- Consolidated poem id isn't useful for joining; what do you think of adding an alt_poem_ids field for additional ids? Then we can privilege one reference id / span but keep the other ids as possible/likely duplicates
- Test cases are probably redundant, especially now that the merge logic is simpler than before

### DIFF
--- a/src/corppa/poetry_detection/merge_excerpts.py
+++ b/src/corppa/poetry_detection/merge_excerpts.py
@@ -18,9 +18,11 @@ Merging logic is as follows:
   combination of `page_id`, `ppa_span_start`, and `ppa_span_end`)
   even when poem identifications differ, and combined as follows:
 
-     - Excerpts are sorted by `poem_id`, and `ref_span_start` with nulls last,
-       and the reference information is taken from the first excerpt (`poem_id`,
-       `ref_span_start`, `ref_span_end`, `ref_span_text`, `ref_corpus`).
+     - Excerpts are sorted by `poem_id`, `ref_span_start`, and passim
+       match length with nulls last and longest passim match first.
+       Reference information (`poem_id`, `ref_span_start`, `ref_span_end`, 
+       `ref_span_text`, `ref_corpus`) is taken from the first excerpt in 
+       the group.
      - When merged excerpts have different poem identifications, all unique
        poem ids after the first are collected into `alt_poem_ids`
      - The `detection_methods` and `identification_methods` fields are combined
@@ -35,7 +37,10 @@ labeled_excerpts.csv -o merged_excerpts.csv
 
 Limitations:
 
-- Currently supports CSV input and output only
+- Merge logic collapses different poem ids that may not correspond;
+  they may be subsets of the same poem, different editions, or entirely
+  different poems.
+- Currently supports CSV input and output only.
 
 """
 
--- a/src/corppa/poetry_detection/merge_excerpts.py
+++ b/src/corppa/poetry_detection/merge_excerpts.py
@@ -105,7 +105,8 @@ def merge_excerpts(
     )
 
     # sort by page then poem id, with nulls last, to ensure we select
-    # a non-null poem id and reference data
+    # a non-null poem id and reference data;
+    # extract passim match length and sort longest passim matches first
     merge_groups = (
         merge_candidates.with_columns(
             # extract passim match length so we can prioritize longer matches
--- a/src/corppa/poetry_detection/merge_excerpts.py
+++ b/src/corppa/poetry_detection/merge_excerpts.py
@@ -139,9 +139,11 @@ def merge_excerpts(
         .drop("group_size")
     )
 
-    merge_groups = merge_candidates.group_by(
-        ["page_id", "ppa_span_start", "ppa_span_end"]
-    )
+    # sort by page then poem id, with nulls last, to ensure we select
+    # a non-null poem id and reference data
+    merge_groups = merge_candidates.sort(
+        "page_id", "poem_id", "ref_span_start", nulls_last=True
+    ).group_by(["page_id", "ppa_span_start", "ppa_span_end"], maintain_order=True)
     num_merge_groups = merge_groups.len().height
     if verbose:
         print(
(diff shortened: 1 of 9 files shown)


src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -132,71 +134,75 @@ def merge_excerpts(
...
+    merged_output_df = (
+        merge_groups.agg(
+            pl.first("ppa_span_text"),  # should match exactly
+            pl.col("detection_methods")
+            .explode()
+            .unique(),  # combine in a single list, no repeats
+            # combine notes but don't repeat duplicate info (like passim char match count)
+            pl.col("notes").explode().unique().sort().str.join("; "),
+            # construct merged excerpt id manually; c= prefix for combined
+            # (although strictly speaking should only be  if > 1 detection method)
```
laurejt: Will we encounter this edge case?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -132,71 +134,75 @@ def merge_excerpts(
...
+            pl.concat_str(
+                pl.lit("c@"),
+                pl.col("ppa_span_start").first(),
+                pl.lit(":"),
+                pl.col("ppa_span_end").first(),
+            ).alias("excerpt_id"),
+            # TODO: better to pick first non-null and put alternates in a separate field?
```
laurejt: Could you select the most frequent (mode) ID?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -132,71 +134,75 @@ def merge_excerpts(
...
+            pl.col("poem_id").explode().unique().sort().str.join("; "),
+            pl.col("ref_corpus").explode().unique().sort().str.join("; "),
+            # use first reference span and text so numbers are useful; ignore nulls
+            pl.col("ref_span_start").drop_nulls().first(),
+            pl.col("ref_span_end").drop_nulls().first(),
+            pl.col("ref_span_text").drop_nulls().first(),
```
laurejt: Since the poem ids and ref corpora are sorted, it's be hard to know which reference span is being displayed

laurejt (approved):
Looks good to me.

I think it makes sense to elevate one poem identification for combined excerpts, but I think we want to have more logic for how this representative is selected. We might want to prioritize particular identification methods or poems. Tracking alternative poem_ids seems like a good idea, but would we also want the span indices for debugging / examining the alternatives?

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: I'm not sure which edge case you mean but I think both of the things mentioned in the comments were happening in my testing.

rlskoeser:
@laurejt updated to sort before grouping to make sure we get a non-null poem id and get all corresponding reference span information for that labeled excerpt. I also made an `alt_poem_ids` field and split out primary id from alternate ids.  I think the sorting could be improved to help choosing the best match, but I can't do that right now.

Here's the output I'm getting when I run the merge script on the current version of the data:


## Merging excerpts
Loaded 1,342,504 excerpts from 6 files (1,342,495 unique; 1,336,763 labeled).
Identified 174,741 merge candidates in 84,320 groups.
84,320 merged excerpts; 82,953 with multiple poem ids.
```
1,252,074 excerpts after merging; 1,247,665 labeled excerpts.
Total by detection method:
        passim: 1,246,136
        adjudication; passim: 71
        xml: 3,233
        passim; xml: 188
        adjudication: 2,446
        manual: 105
        passim: 1,246,262
        manual; passim: 3
        refmatcha: 1,067
        passim; refmatcha: 123
        manual; passim; refmatcha: 7
        manual; refmatcha: 98
```

src/corppa/poetry_detection/merge_excerpts.py
laurejt: In what cases are there exact duplicates for the same detection method?

src/corppa/poetry_detection/merge_excerpts.py
laurejt, line 74: Update the method definition so it matches our new merging approach.

This is especially important to flag our more aggressive merging approach where we merge excerpts with conflicting identified poems.

src/corppa/poetry_detection/merge_excerpts.py
laurejt: Looking at this a bit more, I was confused by the comments themselves and what our working definition for excerpt ids is.

I think I understand the issue now. Since we're now merging conflicting poem identifications, it's possible for two excerpts to be merged that have the same detection method (and at least for passim the same identification method).

Arguably, we could probably fix this discrepancy in the excerpt_id construction by checking the set of detection methods, but it's a small quibble especially if we're rarely using that component of the ID.

src/corppa/poetry_detection/merge_excerpts.py
laurejt: Is this now dead code? I don't see it being called anywhere.

rlskoeser: Thanks for catching! Not needed in the new approach; removed.

rlskoeser: Thank you! Agreed.

rlskoeser:
@laurejt I've removed the dead code, revised the documentation, and added a notebook to explore the merge data a little bit. In doing so I figured out how to prioritize longer passim matches, so I went ahead and implemented that and updated the docs.  (Estimate needs to go up for this work... )

I updated the notebook with a section to look at poem pairs together (updated the poem meta so I can use the same logic for the alt poem id join).  Unfortunately this is hard to share right now because of the state of the data.  (Also unfortunately I have a slight data problem because I didn't know the manual corpus id I pass into passim on the command line becomes the reference corpus id... 😕 )

If you have capacity, would appreciate you looking over the documentation to make sure it's clear.

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -141,9 +106,21 @@ def merge_excerpts(
 
     # sort by page then poem id, with nulls last, to ensure we select
     # a non-null poem id and reference data
```
laurejt: Minor point, but it'd be nice if this comment mentioned the added logic for passim sorting

src/corppa/poetry_detection/polars_utils.py
```diff
@@ -126,13 +126,17 @@ def load_meta_df(file: pathlib.Path, fields_table: dict[str, str]) -> pl.DataFra
 def add_ref_poems_meta(
     excerpts_df: pl.DataFrame,
     ref_poem_meta: pathlib.Path,
+    poem_id_field: str = "poem_id",
+    ref_corpus_field: str = "ref_corpus",
+    suffix: str | None = None,
```
laurejt, line 131: What's this do? Is it some optional arg for polars? Update the docstring to mention what this is for.

laurejt (approved):
This is ready to go after updating a bit of documentation.

PR #257, opened 2026-02-25
Feature/merge near dupes

### LINKED ISSUE #251: write a method to identify near-duplicate poetry excerpts
May be able to adapt some logic from Laure's Homer EoP notebook.

Current merge logic is extremely conservative (exact matches only), and did not merge very many excerpts.  We need to expand this to partial / nearly-complete matches from two different systems. We will need to decide how to combine overlapping spans (most inclusive / largest span, or the smallest span, overlap only?).  We also need to make sure we represent the merge accurately. 

We should start conservatively, since we know that passim tends to grab extra characters around the actual quotes and we don't want to merge based on that. Merge based on high percentage overlap (99% ? ) and require a minimum number of characters.

Start with the manual annotations and match to excerpts from other systems; this will give a more manageable subset to work with and allow working with a known, reliable set of vetted excerpts. Probably useful to work in tandem with #253 to view excerpts before and after merging.

rlskoeser:
**Associated Issue(s):** resolves #251

### Changes in this PR

- Preliminary logic to identify & merge excerpts with a high degree of overlap

### Notes

- Based on previous notebook code and also span overlap & overlap factor
- Does not yet merge excerpts; hoping to figure out a way to refactor & reuse exact span merge logic

### DIFF
--- a/src/corppa/poetry_detection/merge_excerpts.py
+++ b/src/corppa/poetry_detection/merge_excerpts.py
@@ -191,7 +191,82 @@ def merge_excerpts(
     # combined merged records with the output
     # use a diagonal concat instead of vstack/extend
     # to avoid having to reconcile columns first
-    return pl.concat([output_df, merged_output_df], how="diagonal")
+    output_df = pl.concat([output_df, merged_output_df], how="diagonal")
+
+    # now identify & merge partial overlap
+    overlaps_df = (
+        output_df
+        # Filter to excerpts on pages with multiple excerpts
+        .filter(pl.col("page_id").is_duplicated())
+        .join_where(
+            # excerpts_df,
+            output_df,
+            # (1) Excerpts have same page ID
+            pl.col("page_id") == pl.col("page_id_right"),
+            # (2) Excerpts overlap
+            ## Left's start index < Right's end index
+            pl.col("ppa_span_start") < pl.col("ppa_span_end_right"),
+            ## Right's start index < Left's end index
+            pl.col("ppa_span_start_right") < pl.col("ppa_span_end"),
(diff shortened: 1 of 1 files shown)


src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -132,69 +134,148 @@ def merge_excerpts(
...
+        print(
+            f"{merged_output_df.height:,} merged excerpts; {multi_id:,} with multiple poem ids."
+        )
+
+    # combined merged records with the output
+    # use a diagonal concat instead of vstack/extend
+    # to avoid having to reconcile columns first
+    output_df = pl.concat([output_df, merged_output_df], how="diagonal")
+
+    # now identify & merge partial overlap
```
rlskoeser: @laurejt sorry for the confusing diff! I didn't recognize how many changes here. This is the part of the code I wanted your thoughts on

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -202,7 +202,82 @@ def merge_excerpts(
...
+            ## Left's start index < Right's end index
+            pl.col("ppa_span_start") < pl.col("ppa_span_end_right"),
+            ## Right's start index < Left's end index
+            pl.col("ppa_span_start_right") < pl.col("ppa_span_end"),
+            ## Distinct excerpts (excludes self-matches)
+            pl.col("excerpt_id") != pl.col("excerpt_id_right"),
+        )
+        .with_columns(
+            # make a sorted combined id so we can drop duplicate copies of the same pair
+            # (but why doesn't this make the set half the size?)
```
laurejt: doesn't `unique` (L232) do this?

PR #272, opened 2026-04-16
Add method to find overlapping excerpts

### LINKED ISSUE #251: write a method to identify near-duplicate poetry excerpts

rlskoeser:
**Associated Issue(s):** resolves #251 

### Changes in this PR

- Moved logic for merging grouped excerpts into a method with the hope of making it reusable
- Added new method for identifying overlapping excerpts; adapted from existing span overlap factor logic

### Notes
- Overlap length / factor logic is adapted from @laurejt span evaluation code
- Would appreciate input what fields to return and field names (I don't love the `_right` default but haven't come up with anything better); extra fields have been helpful when testing, should they be optional? removed?

### Review checklist
- [x] check logic for identifying high-confidence overlap
- [x] weigh in on returned field question above
- [x] check unit tests - are there important cases I should add? (more than one pair?)


src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+    groups of excerpts into merged excerpts. Merges as follows:
+      - first ppa_span_text
+      - combined unique set of detection methods
+      - combined unique set of notes
+      - updated excerpt id
+      - first poem id (dataframe should be sorted so preferred id is first)
+      - any other poem ids are listed in alt_poem_ids
+      - first reference corpus id
+      - first reference span and text
+      - combined unique list of identification methods
```
laurejt, line 71: This might just be me, but I found this a bit hard to read. It might be helpful describe what "first" means here and show the names of fields being defined in this merge

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+    After merging, it adds a note documenting the group, with the specified reason,
```
laurejt, line 72: Perhaps mention there's a default (`merge_reason`)

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+    and the number of raw excerpts in the merged set.
+    """
+    return (
+        grouped_df.agg(
+            # TODO: how to handle for overlapping spans
+            pl.first("ppa_span_text"),  # should match exactly
+            pl.col("detection_methods")
+            .explode()
+            .unique(),  # combine in a single list, no repeats
```
laurejt: It'd be nice if this could be on one line similar to L81

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+      - first reference corpus id
+      - first reference span and text
+      - combined unique list of identification methods
+    After merging, it adds a note documenting the group, with the specified reason,
+    and the number of raw excerpts in the merged set.
+    """
+    return (
+        grouped_df.agg(
+            # TODO: how to handle for overlapping spans
+            pl.first("ppa_span_text"),  # should match exactly
```
laurejt: Not sure what the comment "should match exactly" means here

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+            pl.col("detection_methods")
+            .explode()
+            .unique(),  # combine in a single list, no repeats
+            # combine notes but don't repeat duplicate info (like passim char match count)
```
laurejt, line 81: how often are there duplicate notes?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+            pl.col("notes").explode().unique().sort().str.join("; "),
```
laurejt: does it make sense to sort these? then we can no longer the "first" note corresponds to the first, representative excerpt

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+            # construct merged excerpt id manually; c= prefix for combined
+            # (although strictly speaking should only be if > 1 detection method)
+            pl.concat_str(
+                pl.lit("c@"),
+                pl.col("ppa_span_start").first(),
+                pl.lit(":"),
+                pl.col("ppa_span_end").first(),
+            ).alias("excerpt_id"),
+            # pick the first poem id (relies on previous sorting)
+            pl.col("poem_id").explode().unique().first(),
```
laurejt: Isn't the poem_id field singular?
```suggestion
            pl.col("poem_id").first(),
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+def merge_excerpt_groups(
+    grouped_df: pl.dataframe.group_by.GroupBy, merge_reason: str = "ppa exact span"
+) -> pl.DataFrame:
+    """Takes a GroupBy dataframe of excerpts (created by calling `group_by`), and combines
+    groups of excerpts into merged excerpts. Merges as follows:
+      - first ppa_span_text
+      - combined unique set of detection methods
+      - combined unique set of notes
+      - updated excerpt id
+      - first poem id (dataframe should be sorted so preferred id is first)
```
laurejt, line 67: What do you mean "be sorted"?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+            ).alias("excerpt_id"),
+            # pick the first poem id (relies on previous sorting)
+            pl.col("poem_id").explode().unique().first(),
+            # and store all others in alt poem ids field
+            pl.col("poem_id")
+            .explode()
+            .unique()
+            .drop_nulls()
+            .slice(1)
+            .alias("alt_poem_ids"),
```
laurejt: Wouldn't something like this work? It feels like we might want to keep a null around as an indicator that there is at least one unidentified excerpt in the pool
```suggestion
            pl.col("poem_id")
            .tail(-1)
            .unique()
            .alias("alt_poem_ids"),
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+            pl.col("ref_corpus").explode().first(),
```
laurejt:
```suggestion
            pl.col("ref_corpus").first(),
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+            .unique(),  # combine in a single list, no repeats
+            # combine notes but don't repeat duplicate info (like passim char match count)
+            pl.col("notes").explode().unique().sort().str.join("; "),
+            # construct merged excerpt id manually; c= prefix for combined
+            # (although strictly speaking should only be if > 1 detection method)
+            pl.concat_str(
+                pl.lit("c@"),
+                pl.col("ppa_span_start").first(),
+                pl.lit(":"),
+                pl.col("ppa_span_end").first(),
```
laurejt: Is it worth revisiting our ID structure?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+            pl.col("ref_corpus").explode().first(),
+            # use first reference span and text so numbers are useful; ignore nulls
+            pl.col("ref_span_start").first(),
+            pl.col("ref_span_end").first(),
+            pl.col("ref_span_text").first(),
+            # combine unique list of id methods
+            pl.col("identification_methods")
+            .explode()
+            .unique()
+            .drop_nulls(),  # combine in a single list, no repeats, ignore nulls (not identified before merging)
```
laurejt: It'd be nice if this was on one line for readability

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+            pl.len().alias("group_size"),  # count number in the group
```
laurejt, line 102: This is computed in multiple places, maybe it's worth keeping this additional field for development/debugging?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -54,6 +54,77 @@
...
+            # if alt poem ids is empty, replace with None
+            alt_poem_ids=pl.when(pl.col("alt_poem_ids").list.len() > 0)
+            .then(pl.col("alt_poem_ids"))
+            .otherwise(pl.lit(None)),
+        )
+        .drop("group_size")
+    )  # drop group size column
+
+
 def merge_excerpts(
```
laurejt, line 120: Update the description if merge behavior has changed

src/corppa/poetry_detection/merge_excerpts.py
laurejt, line 145: Is this still needed?

laurejt, line 151: This field is also create / used by `merge_excerpt_groups`

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -191,7 +213,102 @@ def merge_excerpts(
...
+    min_overlap_chars: int = 10,
+) -> pl.DataFrame:
+    """
+    Takes a DataFrame of excerpts and identifies pairs of overlapping excerpts.
+    Overlapping excerpts are on the same page, with some shared span of text.
+    We exclude short overlaps based on the minimum character parameter,
+    and an overlap factor, which is calculated by the length of the shared
+    span divided by the length of the longer of the two spans.  Note that
+    this will typically not return small excerpts completely inside another
+    larger span.
```
laurejt, line 223: Should full subset excerpts be handled differently? It's an easy enough edge case

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -191,7 +213,102 @@ def merge_excerpts(
...
+    """
+
+    # identify excerpts with partial overlap
+    overlaps_df = (
+        excerpts_df
+        # Filter to excerpts on pages with multiple excerpts
+        .filter(pl.col("page_id").is_duplicated())
+        .join_where(
+            excerpts_df,
+            # 1. Limit to excerpts are on the same page
```
laurejt:
```suggestion
            # 1. Limit to excerpts on the same page
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -191,7 +213,102 @@ def merge_excerpts(
...
+            pl.col("ppa_span_start_right") < pl.col("ppa_span_end"),
+            # 3. Exclude self-matches
+            pl.col("excerpt_id") != pl.col("excerpt_id_right"),
+        )
+        .with_columns(
+            # make a sorted combined id so we can drop duplicate copies of the same pair
+            group_ids=pl.concat_list(
+                [pl.col("excerpt_id"), pl.col("excerpt_id_right")]
+            ).list.sort()
+        )
```
laurejt: We should be able to avoid this duplication by using an additional conditional. Perhaps ordering of excerpt IDs (i.e., excerpt_id  < excerpt_id_right)

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -191,7 +213,102 @@ def merge_excerpts(
...
+            excerpts_df,
+            # 1. Limit to excerpts are on the same page
+            pl.col("page_id") == pl.col("page_id_right"),
+            # 2. Excerpts overlap:
+            #    left span starts before right span ends
+            pl.col("ppa_span_start") < pl.col("ppa_span_end_right"),
+            #    and right span starts before left span ends
+            pl.col("ppa_span_start_right") < pl.col("ppa_span_end"),
+            # 3. Exclude self-matches
+            pl.col("excerpt_id") != pl.col("excerpt_id_right"),
```
laurejt: Not sure polars can do this, but if we have an ordering for excerpt_ids then, we could avoid duplicating merges.

```suggestion
            pl.col("excerpt_id") < pl.col("excerpt_id_right"),
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -191,7 +213,102 @@ def merge_excerpts(
...
+                    pl.col("ppa_span_end_right").sub(pl.col("ppa_span_start_right")),
+                    # pl.col("ppa_span_text").str.len_chars(),
+                    # pl.col("ppa_span_text_right").str.len_chars(),
+                )
+            )
+        )
+        # filter to requested overlap / length to limit to high confidence overlaps
+        .filter(
+            pl.col("overlap_factor").ge(min_overlap_factor),
+            pl.col("overlap_len").ge(min_overlap_chars),
```
laurejt: We could do some additional filtering by skipping excerpts shorter than minimum overlap

laurejt (commented):
I've added my initial thoughts on `merge_excerpts.py`, but haven't looked at its unit tests yet.

Some higher-level musings:
- Perhaps it'd be useful to add an additional field with excerpt length given how often we operate on it.
- Would it ever be useful to use the `map_groups` method? I know it's generally slower, but perhaps it could be useful at some point...

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: Possible, but I don't think it matters enough to add (additional complexity, and this is already pretty fast).

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: Merging them is easy but identifying the correct one is hard. Out of scope for current effort in either case.

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: Possibly at some point, but out of scope for current work. I'm not sure how meaningful part is, but I don't think it's hurting anything.

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: Keeping the nulls isn't really meaningful, it's an artifact of the detection methods not having an identification component (since only passim does both). That information is more meaningfully tracked by the methods.

rlskoeser:
@laurejt I've cleaned up based on your feedback and added the core objects to our sphinx docs (which I also split out into separate files for each top-level module to be more manageable).    There are more changes in the PR now because I've cleaned up the docs, so please focus on the `identify_overlapping_excerpts` method: do you see any problems with the logic? Do you have an opinion on the fields returned? Are there any important cases you think I should cover in the unit tests?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -58,67 +58,62 @@ def merge_excerpt_groups(
...
-      - updated excerpt id
-      - first poem id (dataframe should be sorted so preferred id is first)
-      - any other poem ids are listed in alt_poem_ids
-      - first reference corpus id
-      - first reference span and text
-      - combined unique list of identification methods
-    After merging, it adds a note documenting the group, with the specified reason,
-    and the number of raw excerpts in the merged set.
+    groups of excerpts into merged excerpts.  Fields are expected to correspond to
+    labeled excerpts (:class:`~corppa.poetry_detection.core.LabeledExcerpt`), and the
```
laurejt, line 62: Is it possible for there to be a group of unlabeled excerpts that should be merged (i.e., the collective `poem_id` would be null)?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -58,67 +58,62 @@ def merge_excerpt_groups(
...
     """
     return (
         grouped_df.agg(
-            # TODO: how to handle for overlapping spans
-            pl.first("ppa_span_text"),  # should match exactly
-            pl.col("detection_methods")
-            .explode()
-            .unique(),  # combine in a single list, no repeats
+            # NOTE: does not handle overlapping spans;
+            # currently assumes text spans match or first in group is complete
```
laurejt, line 81: Does "complete" here mean is labeled?

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -49,37 +48,95 @@
...
+                c.ppa_span_start.first(),
+                pl.lit(":"),
+                c.ppa_span_end.first(),
+            ).alias("excerpt_id"),
+            # pick the first poem id (relies on previous sorting)
+            c.poem_id.first(),
+            # and store all others in alt poem ids field
+            c.poem_id.unique().drop_nulls().slice(1).alias("alt_poem_ids"),
+            c.ref_corpus.first(),
+            # use first reference span and text so numbers are useful; ignore nulls
```
laurejt:
```suggestion
            # use first reference span and text so numbers are useful
```

src/corppa/poetry_detection/merge_excerpts.py
```diff
@@ -128,70 +186,97 @@ def merge_excerpts(
...
+            )
+        )
+        # filter to requested overlap / length to limit to high confidence overlaps
+        .filter(
+            c.overlap_factor.ge(min_overlap_factor),
+            c.overlap_len.ge(min_overlap_chars),
+        )
+    )
+
+    # what fields are needed here?
```
laurejt, line 261: Including poem ids (when available) could also be useful

tests/test_poetry_detection/test_merge_excerpts.py
laurejt, line 91: I think it's worth including a test that checks for the expected behavior when the input excerpts are in a different ordering

laurejt, line 1: Does "ref id" correspond to "poem id", if so I think it's worth only using the latter term

tests/test_poetry_detection/test_merge_excerpts.py
```diff
@@ -507,3 +509,100 @@ def test_main_successful(capsys, tmp_path):
     assert merged_ex1.ref_span_text == excerpt1_label1.ref_span_text
     # id methods combined
     assert merged_ex1.identification_methods == {"manual", "refmatcha"}
+
+
+### test for identify_overlapping_excerpts
```
laurejt, line 526: I think it's worth adding a couple unit tests for testing the behavior of specifying `min_overlap_factor` and `min_overlap_chars` especially the boundaries (i.e., greater than vs greater)

tests/test_poetry_detection/test_merge_excerpts.py
```diff
@@ -507,3 +509,100 @@ def test_main_successful(capsys, tmp_path):
...
+
+# test scenarios that should result in no overlapping pairs
+no_overlap_inputs = [
```
laurejt, line 529: I think it's worth adding two additional cases testing that both overlap constraints need to be met

- Excerpts with perfect overlap, but are too small
- Excerpts with a long overlap (i.e., > `min_overlap_chars`), but too low overlap factor

I realized it's done in a later unit test, but it feels nicer organizationally to have it here for the default settings

tests/test_poetry_detection/test_merge_excerpts.py
```diff
@@ -507,3 +509,100 @@ def test_main_successful(capsys, tmp_path):
...
+def test_identify_overlapping_excerpts_no_pairs(excerpts, reason):
+    # construct a standardized dataframe from the list of excerpts given
+    excerpts_df = standardize_dataframe(
+        pl.from_dicts([ex.to_dict() for ex in excerpts])
+    )
+    pairs_df = identify_overlapping_excerpts(excerpts_df)
+    assert pairs_df.height == 0, f"expected 0 overlapping pairs: {reason}"
+
+
+def test_identify_overlapping_excerpts():
```
laurejt, line 578: I think it's worth adding an additional unit test to check the output ordering since there currently is an assumed one in the code. Also putting into the method docstrings because it's a useful behavior to know about when looking for all possible overlaps of a specific excerpt

laurejt (approved):
This is looking really good. I think it'd be nice to include poem id fields for `identify_overlapping_excerpts`'s output. See my comments for some additional unit tests that you should consider adding primarily around the ordering logic of results.

laurejt:
@rlskoeser When I build the docs locally, I receive the following warning:

> /Users/lauret/cdh-dev/ppa/corppa/README.md:36: WARNING: 'myst' cross-reference target not found: 'LICENSE' [myst.xref_missing]

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: Possible but not very likely; I don't think any of the logic here would cause problems in that case. If you want to improve the language please suggest changes.

src/corppa/poetry_detection/merge_excerpts.py
rlskoeser: I think what I meant is that it is the full text / longest span of the group but I agree it doesn't make a lot of sense - right now we're only using this on equal spans and I do have a note about that

src/corppa/poetry_detection/merge_excerpts.py
laurejt: I think it's fine, just figured I'd ask. It's just a slight type misalignment since a `LabeledExcerpt`'s `poem_id` field cannot be `None`

rlskoeser: right, technically it could be a mix of labeled and unlabeled excerpts

rlskoeser:
Added missing target for license in the docs, added additional test cases.

